PR #19, opened 2018-11-30
Re-export Skycfg modules as Stalark predeclared values.

dilyevsky:
This would allow users to use Skycfg modules as additional predeclared packages in their own Starlark runtime.


skycfg.go
```diff
@@ -149,25 +149,42 @@ func WithProtoRegistry(r unstableProtoRegistry) LoadOption {
...
-			"url":    impl.UrlModule(),
-		},
-		fileReader: LocalFileReader(filepath.Dir(filename)),
-	}
+	parsedOpts := &loadOptions{}
 	for _, opt := range opts {
 		opt.applyLoad(parsedOpts)
 	}
-	protoModule.Registry = parsedOpts.protoRegistry
+	parsedOpts.globals = UnstablePredeclaredModules(parsedOpts.protoRegistry)
```
jmillikin-stripe: This will prevent the `WithGlobals()` option from working, because its inserted values will be overwritten. Same issue for `WithFileReader()`.

It's a little verbose, but maybe something like this would work for now?

```go
func UnstablePredeclaredModules(r unstableProtoRegistry) starlark.StringDict {
	modules, protoModule := predeclaredModules()
	protoModule.Registry = r
	return modules
}

func predeclaredModules() (starlark.StringDict, *impl.ProtoModule) { ... }
```

Then `Load()` would still have access to the internal proto module type and could set its registry directly.

dilyevsky: Ah, sorry. What do you think about this instead?

skycfg.go
```diff
@@ -149,25 +149,48 @@ func WithProtoRegistry(r unstableProtoRegistry) LoadOption {
...
 	}
+
 	for _, opt := range opts {
 		opt.applyLoad(parsedOpts)
 	}
-	protoModule.Registry = parsedOpts.protoRegistry
+
+	// Set the proto registry option if proto package was configured.
+	if p, ok := parsedOpts.globals["proto"]; ok {
+		p.(*impl.ProtoModule).Registry = parsedOpts.protoRegistry
```
jmillikin: This will crash if the user set `proto` to something else.

dilyevsky: Hm, that's right - one could override `proto` key...

```
 WithGlobals adds *additional* global symbols
```

^ should we allow overrides though?

jmillikin-stripe: We currently allow overrides to provide users an escape hatch, and for stubbing in tests. Even if we decide to forbid it, the user-visible behavior should be an error message instead of a type-conversion panic.

dilyevsky: > Even if we decide to forbid it, the user-visible behavior should be an error message instead of a type-conversion panic.

Yeah, totally. Was just thinking what's the best way to tackle this.

jmillikin-stripe: If constructing the default globals is moved into a helper, then we can safely provide both the new interface (returning a single dict) and support the existing `Load()` semantics.

dilyevsky:
Sorry forgot about this one. Addressed original concerns. PTAL

Issue #30, opened 2019-01-07
Syntax/required boilerplate for accessing a top-level enum?


pcn:
I've been successful at accessing nested enums in protobuf messages, but it seems like the top-level enum at https://github.com/envoyproxy/envoy/blob/master/api/envoy/api/v2/core/base.proto#L122 requires something more than `package.name.enum_name.SYMBOL`

My test case is this, based mainly off of `_examples/repl`:

`main.go`
```go
package main

import (
	"context"
	"fmt"
	"reflect"

	docopt "github.com/docopt/docopt-go"

	"github.com/gogo/protobuf/jsonpb"
	"github.com/gogo/protobuf/proto"
	_ "github.com/golang/protobuf/ptypes/wrappers"
	"github.com/stripe/skycfg"
	yaml "gopkg.in/yaml.v2"

	gogo_proto "github.com/gogo/protobuf/proto"

	_ "github.com/envoyproxy/go-control-plane/envoy/admin/v2alpha"
	_ "github.com/envoyproxy/go-control-plane/envoy/api/cluster"
	_ "github.com/envoyproxy/go-control-plane/envoy/api/v2"
	_ "github.com/envoyproxy/go-control-plane/envoy/config/bootstrap/v2"
)

type protoRegistry struct{}

func (*protoRegistry) UnstableProtoMessageType(name string) (reflect.Type, error) {
	if t := proto.MessageType(name); t != nil {
		return t, nil
	}
	if t := gogo_proto.MessageType(name); t != nil {
		return t, nil
	}
	return nil, nil
}

func (*protoRegistry) UnstableEnumValueMap(name string) map[string]int32 {
	if ev := proto.EnumValueMap(name); ev != nil {
		return ev
	}
	if ev := gogo_proto.EnumValueMap(name); ev != nil {
		return ev
	}
	return nil
}

func main() {

	usage := `envoy-proto: creates an envoy configuration from a skycfg file

Usage:
  envoy-proto <input-file>
  envoy-proto -h | --help

Options:
  <input-file>  skycfg file to be read
  -h --help     Show this screen.`

	arguments, _ := docopt.ParseDoc(usage)

	ctx := context.Background()
	config, err := skycfg.Load(ctx, arguments["<input-file>"].(string), skycfg.WithProtoRegistry(&protoRegistry{}))
	if err != nil {
		panic(err)
	}
	messages, err := config.Main(ctx)
	if err != nil {
		panic(err)
	}
	for _, msg := range messages {
[...]

PR #31, opened 2019-01-07
Use custom proto registry for proto.package.

dilyevsky:
Use custom registry to lookup enum names in proto package attribute call (instead of standard `proto.EnumValueMap`). This should fix issues with top-level enums (#30) when protobuf registry is overridden in the proto package (e.g gogo proto).


pcn:
That seems to work for me. Thanks!

PR #35, opened 2019-01-09
Support gogoproto.stdduration option.

### LINKED ISSUE #32: Another envoy-related question - duration bottoming out as "time".Duration
Trying to assign to an `envoy.api.v2.Cluster.connect_timeout` seems to end up tripping over this definition:

https://github.com/envoyproxy/envoy/blob/master/api/envoy/api/v2/cds.proto#L114-L118

which seems to translate to https://github.com/envoyproxy/go-control-plane/blob/master/envoy/api/v2/cds.pb.go#L261

I guess this is another issue with gogo, but I'm not much of a gopher, so hopefully this helps.

I'm using the same main.go as in my prior issues, here is the sky file:

```python
pb = proto.package("google.protobuf")
v2 = proto.package("envoy.api.v2")

def duration():
    return v2.Cluster(
        lb_policy=v2.Cluster.LbPolicy.ROUND_ROBIN,  # RR is the default, but let's be explicit
        type=v2.Cluster.DiscoveryType.STRICT_DNS,
        connect_timeout=pb.Duration(seconds=0, nanos=500000000),  #0.5 seconds
        name="no_duration")
     
duration()  
```

The output is this:

```
./envoy-proto no_duration.sky
panic: TypeError: value <google.protobuf.Duration nanos:500000000 > (type `google.protobuf.Duration') can't be assigned to type `"time".Duration'.

goroutine 1 [running]:
main.main()
	/home/spacey/go/src/github.com/pcn/envoy-proto/main.go:69 +0x6f3
```

dilyevsky:
* Support `gogoproto.stdduration` gogo proto option that generates plain `time.Duration` as proto field via implicit conversion from gogo's `google.protobuf.Duration` proto.

Example usage:
```python
v2 = proto.package("envoy.api.v2")
pb = proto.package("google.protobuf.Duration")

cluster = v2.Cluster(connect_timeout = pb.Duration(seconds = 1))
```

Fixes #32


internal/go/skycfg/time.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+}
+
+func fnDuration(_ *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var s string
+	if err := starlark.UnpackPositionalArgs(b.Name(), args, kwargs, 1, &s); err != nil {
+		return nil, err
+	}
+
+	d, err := time.ParseDuration(s)
```
jmillikin-stripe: I'm not sure about exposing the Go duration syntax directly in the skycfg API, because it's difficult to implement in other versions of the Starlark language (Rust, Java).

Wouldn't `connect_timeout = Duration(seconds = 1)` be fine?

dilyevsky: > I'm not sure about exposing the Go duration syntax directly in the skycfg API, because it's difficult to implement in other versions of the Starlark language (Rust, Java).

Yeah, not sure either. Another option is to do implicit conversion to/from `google.protobuf.Duration` which seems very un-intuitive (to the user)...

> Wouldn't connect_timeout = Duration(seconds = 1) be fine, using the protobuf duration message directly?

This would require special-casing in `valueFromStarlark` to convert from `google.protobuf.Duration` (wrapped into `skyProtoMessage`) to plain `time.Duration`.

I can go ahead an change it if we are settled that implicit conversion is the way to go. lmk

jmillikin-stripe: I'd like to hide the gogo-proto weirdness as much as possible. These fields _should_ be a `*duration_pb.Duration` in the generated Go, and it's only gogo's bizarro options that are turning it into some other type.

dilyevsky:
PTAL

internal/go/skycfg/proto_message.go
```diff
@@ -23,7 +23,9 @@ import (
 	"reflect"
 	"sort"
 	"strings"
+	"time"
 
+	"github.com/gogo/protobuf/types"
```
jmillikin-stripe: Lets use the standard go-protobuf here so we don't need to drag in a hard dependency on GoGo: https://godoc.org/github.com/golang/protobuf/ptypes#DurationProto

internal/go/skycfg/proto_message.go
```diff
@@ -357,6 +361,16 @@ func valueFromStarlark(t reflect.Type, sky starlark.Value) (reflect.Value, error
 			val.Elem().Set(reflect.ValueOf(sky.msg).Elem())
 			return val.Elem(), nil
 		}
+
+		dpb, ok := sky.msg.(*types.Duration)
+		if ok && t == reflect.TypeOf(time.Duration(0)) {
+			d, err := types.DurationFromProto(dpb)
+			if err != nil {
+				return reflect.Value{}, fmt.Errorf("TypeError: %v (type `%s') can't be coverted to type `%s'.", dpb, reflect.TypeOf(dpb), reflect.TypeOf(t))
```
jmillikin-stripe: This would be a `ValueError`, not `TypeError`. The types match, but the value is out of range or invalid. I think you'll also want to include the error returned by [`ptypes.Duration()`](https://godoc.org/github.com/golang/protobuf/ptypes#Duration), it will tell the user what went wrong.

```go
d, err := ptypes.Duration(dpb)
if err != nil {
  return reflect.Value{}, fmt.Errorf("ValueError: %v", err)
}
```

dilyevsky: Ah yes, good catch, thanks.

internal/go/skycfg/proto_message.go
dilyevsky: Hm, this would mean user must make sure to return `google.protobuf.Duration`-reflected type from the standard registry, not gogo (which they are probably going to be using since it's a gogo-only option).

jmillikin-stripe: If I have skycfg code like this:

```python
envoy = proto.package("envoy.api.v2")
pb = proto.package("google.protobuf")

cluster = v2.Cluster(connect_timeout = pb.Duration(seconds = 1))
```

Then the `pb.Duration` is probably being resolved to the go-protobuf implementation. It will be confusing if we need to add a new lookup function or option to make `pb.Duration()` specifically return the GoGo version.

Similarly, `cluster.connect_timeout` ought to return the normal types instead of the GoGo types so that users don't have to contend with:

```python
ct = cluster.connect_timeout
cluster.connect_timeout = ct # TypeError, wtf?
```

The part we can't solve, and this is part of why the existence of GoGo is so frustrating, is that GoGo-generated code can have different duration types even in the same message. I don't think there's anything in Envoy or Kubernetes that does such a thing, so we can worry about it later.

dilyevsky: > Then the pb.Duration is probably being resolved to the go-protobuf implementation.

That actually will depends on which registry does the lookup first =) so one would need to make sure they implemented their custom registry as follows (or just use the default registry):

```
func (*protoRegistry) UnstableProtoMessageType(name string) (reflect.Type, error) {
	if t := proto.MessageType(name); t != nil {
		return t, nil
	}
	if t := gogo_proto.MessageType(name); t != nil {
		return t, nil
	}
	return nil, nil
}
```

dilyevsky: Done.

PR #36, opened 2019-01-16
Add function to execute tests

choo-stripe:
In addition to writing config generation in skycfg, users need to test
the config generators. This PR adds a function that will execute tests.
Tests should be named of the form `test_...()` and fail using 
`fail("some msg")`, or return None to pass.


choo-stripe:
Thinking more about this, I want to integrate `assert` and maybe junit xml output too. Closing for now, I'll open another PR.

PR #37, opened 2019-01-17
Adding test infrastructure to skycfg

choo-stripe:
We need a way of executing tests in skycfg, beyond just running `main`. Here's an implementation, open to suggestions.

### DIFF
--- a/skycfg.go
+++ b/skycfg.go
@@ -221,9 +221,12 @@ func loadImpl(ctx context.Context, opts *loadOptions, filename string) (starlark
 		cache[modulePath] = &cacheEntry{globals, err}
 
 		for name, val := range globals {
-			if strings.HasPrefix(name, "test_") && val.Type() == "function" {
+			if !strings.HasPrefix(name, "test_") {
+				continue
+			}
+			if fn, ok := val.(starlark.Callable); ok {
 				tests = append(tests, &Test{
-					callable: val.(starlark.Callable),
+					callable: fn,
 				})
 			}
 		}
@@ -330,7 +333,7 @@ func (c *Config) Main(ctx context.Context, opts ...ExecOption) ([]proto.Message,
 
 // A TestResult is the result of a test run
 type TestResult struct {
-	Name     string
+	TestName string
 	Failure  error
 	Duration time.Duration
 }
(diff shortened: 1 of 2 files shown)


skycfg.go
```diff
@@ -215,12 +219,21 @@ func loadImpl(ctx context.Context, opts *loadOptions, filename string) (starlark
 		cache[modulePath] = nil
 		globals, err := starlark.ExecFile(thread, modulePath, moduleSource, opts.globals)
 		cache[modulePath] = &cacheEntry{globals, err}
+
+		for name, val := range globals {
+			if strings.HasPrefix(name, "test_") && val.Type() == "function" {
```
jmillikin-stripe: I think we'll want to use a Go type cast here, so we don't need to make assumptions about Starlark type names.

```go
if !strings.HasPrefix(name, "test_") {
    continue
}
if fn, ok := val.(starlark.Callable); ok {
    tests = append(tests, &Test{
        callable: fn,
    })
}

```

skycfg.go
```diff
@@ -315,6 +328,68 @@ func (c *Config) Main(ctx context.Context, opts ...ExecOption) ([]proto.Message,
 	return msgs, nil
 }
 
+// A TestResult is the result of a test run
+type TestResult struct {
+	Name     string
```
jmillikin-stripe: nit: `Name` -> `TestName`

skycfg.go
```diff
@@ -328,3 +403,37 @@ func skyFail(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwar
 	t.Caller().WriteBacktrace(&buf)
 	return nil, fmt.Errorf("[%s] %s\n%s", t.Caller().Position(), msg, buf.String())
 }
+
+// AssertionError represents a failed assertion
+type AssertionError struct {
```
jmillikin-stripe: Lets name this `assertionError` so it doesn't clutter up in the godoc output.

skycfg.go
choo-stripe: Good call.

choo-stripe:
@jmillikin-stripe updated :)

PR #39, opened 2019-01-25
Add assertion functions to testing contexts

choo-stripe:


internal/go/skycfg/assert.go
```diff
@@ -0,0 +1,145 @@
+// Copyright 2018 The Skycfg Authors.
```
jmillikin-stripe: 2018 -> 2019, here and in the other new file.

internal/go/skycfg/assert.go
```diff
@@ -0,0 +1,145 @@
...
+	"go.starlark.net/starlark"
+	"go.starlark.net/syntax"
+)
+
+// AssertModule contains assertion functions.
+// The *TestContext returned can be used to track assertion failures.
+// assert.* functions from this module will mutate the *TestContext.
+// After execution is complete, TestContext.Failures will be non-empty
+// if any of the assertions failed, and also contain details about the failures.
+func AssertModule() (*TestContext, starlark.Value) {
```
jmillikin-stripe: If you have `*TestContext` implement the `starlark.Value` interface, then it'll be self-contained and easier to understand where the state is.

internal/go/skycfg/assert.go
```diff
@@ -0,0 +1,145 @@
...
+	ctx := &TestContext{}
+	return ctx, &Module{
+		Name: "assert",
+		Attrs: starlark.StringDict{
+			"true": starlark.NewBuiltin("assert.true", ctx.AssertUnaryImpl),
```
jmillikin-stripe: If `*TestContext` implements `starlark.Callable`, then you can have test code like this:

```python
def test_something(t):
  t.assert(True)
  t.assert.equal(1 + 2, 3)
```

internal/go/skycfg/assert.go
```diff
@@ -0,0 +1,145 @@
...
+
+			// names match https://github.com/google/starlark-go/blob/7b3aad4436b8cbd25fda2bd658ed44b3dc2c6dcc/syntax/scan.go#L64
```
jmillikin-stripe: Lets use more expressive names here. The short enum identifiers used when parsing Starlark aren't relevant to most users of Skycfg as a configuration library.

```python
assert.equal
assert.not_equal
assert.lesser
assert.greater
assert.lesser_or_equal
assert.greater_or_equal
```

PR #42, opened 2019-01-29
Give the caller a hint that the `gogocompat` module exists

pcn:
On skycfg typeErrors where the source and destination type's
stringifcation look identical, we may have an issue with shadowed
protobuf message names due to how gogo_proto works. If a project
like envoy has chosen gogo, this will happen to all fundamental
message types (time durations, int32 uint32, etc. etc. etc.).

Tell the user this problem exists so the user doesn't have to
go through the ritual cleansing of an embarrassing trial by fire,
and the maintainers don't have to answer this question.

Also, this functions as a placeholder until this stuff has
stabilized and is documented.

### DIFF
--- a/internal/go/skycfg/proto_message.go
+++ b/internal/go/skycfg/proto_message.go
@@ -538,21 +538,13 @@ func typeName(t reflect.Type) string {
 }
 
 func typeError(t reflect.Type, sky starlark.Value) error {
-	if sky.Type() == typeName(t) {
-		return fmt.Errorf(
-			"TypeError: value %s (type `%s') can't be assigned to type `%s'."+
-				"\nThe names of the types in this error are deceptively similar.\n"+
-				"this is often because of a conflict between the google protobuf\n"+
-				"implementation and messages that are shadowing the google naming,\n"+
-				"but which are actually coming from gogoproto, which adds features\n"+
-				"with the unfortunate effect of sometimes causing this confusion.\n\n"+
-				"To address this, try adding the github.com/stripe/skycfg/gogocompat\n"+
-				"module, and follow the examples of the added diffs at\n"+
-				"https://github.com/stripe/skycfg/pull/41/files\n",
+	if sky.Type() != typeName(t) {
+		return fmt.Errorf("TypeError: value %s (type `%s') can't be assigned to type `%s'.",
 			sky.String(), sky.Type(), typeName(t))
-	} else {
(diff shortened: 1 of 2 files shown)


internal/go/skycfg/proto_message.go
```diff
@@ -538,7 +538,21 @@ func typeName(t reflect.Type) string {
 }
 
 func typeError(t reflect.Type, sky starlark.Value) error {
-	return fmt.Errorf("TypeError: value %s (type `%s') can't be assigned to type `%s'.", sky.String(), sky.Type(), typeName(t))
+	if sky.Type() == typeName(t) {
```
jmillikin-stripe: Go style nit: this would be a bit easier to read if it's unindented, and if the multi-line error used a raw string literal.

```go
func typeError(t reflect.Type, sky starlark.Value) error {
	if sky.Type() != typeName(t) {
		return fmt.Errorf("TypeError: value %s (type '%s') can't be assigned to type '%s'.", sky.String(), sky.Type(), typeName(t))
	}
	return fmt.Errorf(`TypeError: value %s (type '%s') can't be assigned to type '%s'.
etc`,
	sky.String(), sky.Type(), typeName(t))
}
```

internal/go/skycfg/proto_message.go
```diff
@@ -538,7 +538,21 @@ func typeName(t reflect.Type) string {
...
+		return fmt.Errorf(
+			"TypeError: value %s (type `%s') can't be assigned to type `%s'."+
```
jmillikin-stripe: I'm slightly -1 on this message as is -- it's a lot of text, and it's essentially asking the user to work around a point of poor usability in the library.

Do you think the details could be put into the README, and the error could point people there?

pcn: Sure, I think that's doable.

internal/go/skycfg/proto_message.go
pcn: Sure, will do.

pcn: I want the top of the error to match. The error has backticks, which get ugly with raw strings, so my (maybe mistaken) feeling is to go with string concatenation again.

jmillikin-stripe: That's fine. The shorter message kept it readable.

jmillikin-stripe:
All merged in. Thank you for your contribution!

Issue #45, opened 2019-02-19
Using skycfg with proto.Any


michaelkipper:
Related to #14, the `generated.proto` files in the Kubernetes API contain the metadata. (E.g.: https://github.com/kubernetes/api/blob/master/apps/v1/generated.proto#L221-L233)

Some hackery can extract the GroupVersionKind from the message type, which sucks but is maintainable.

But other Kubernetes resources (e.g. the Istio `VirtualService` @ https://github.com/istio/api/blob/master/networking/v1alpha3/virtual_service.proto#L119) don't have the `ObjectMeta` embedded. They rely on other constructs like `client-gen` to generate the Go struct but never generate the full protobuf message (E.g.: https://github.com/michaelkipper/istio-client-go/blob/master/pkg/apis/networking/v1alpha3/virtual_service.go#L33-L38)

One thought I had was to generate my own Protobuf wrapper:

```proto
message KubernetesWrapper {
  string APIVersion = 1 [json_name="apiVersion"];
  string Kind = 2 [json_name="kind"];
  k8s.io.apimachinery.pkg.apis.meta.v1.ObjectMeta ObjectMeta = 3 [json_name="metadata"];
  google.protobuf.Any Spec = 4 [json_name="spec"];
}
```

That way we could localize the hackery in our `.skylib` files, rather than in the parser.

The question is, how do we get skycfg to process the `google.protobuf.Any` message?
Happy to contribute this upstream, if someone can help out with the design.

jmillikin-stripe:
Closing: support for `google.protobuf.Any` was implemented in PR https://github.com/stripe/skycfg/pull/83

PR #46, opened 2019-02-20
Add support for proto.to_any

michaelkipper:
Re: https://github.com/stripe/skycfg/issues/45

This simple PR adds support for a `proto.to_any()` function, so we can create schema agnostic protobuf messages with Skycfg.

### DIFF
--- a/internal/go/skycfg/proto_api.go
+++ b/internal/go/skycfg/proto_api.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/ptypes"
 	"go.starlark.net/starlark"
 	yaml "gopkg.in/yaml.v2"
 )
@@ -53,6 +54,7 @@ func NewProtoModule(registry ProtoRegistry) *ProtoModule {
 			"merge":        starlark.NewBuiltin("proto.merge", fnProtoMerge),
 			"set_defaults": starlark.NewBuiltin("proto.set_defaults", fnProtoSetDefaults),
 			"to_json":      starlark.NewBuiltin("proto.to_json", fnProtoToJson),
+			"to_any":       starlark.NewBuiltin("proto.to_any", fnProtoToAny),
 			"to_text":      starlark.NewBuiltin("proto.to_text", fnProtoToText),
 			"to_yaml":      starlark.NewBuiltin("proto.to_yaml", fnProtoToYaml),
 		},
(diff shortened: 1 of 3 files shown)


michaelkipper:
@jmillikin-stripe: What do you think about this PR?

internal/go/skycfg/proto_api.go
```diff
@@ -234,6 +238,33 @@ func fnProtoToJson(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple
...
+	// Disambiguate between golang and gogo encoded proto messages.
+	var any proto.Message
+	if "" != proto.MessageName(msg.msg) {
+		// Returns a golang any.Any type.
+		any, err = ptypes.MarshalAny(msg.msg)
+	} else if "" != gogo_proto.MessageName(msg.msg) {
+		// Returns a gogo types.Any type.
+		any, err = gogo_types.MarshalAny(msg.msg)
+	} else {
+		return nil, fmt.Errorf("%s: could not get message name for %s", "proto.to_any", msg.Type())
```
dilyevsky:
```suggestion
		return nil, fmt.Errorf("%s: could not get message name for %s", fn.Name(), msg.Type())
```

internal/go/skycfg/proto_api.go
```diff
@@ -234,6 +238,33 @@ func fnProtoToJson(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple
 	return starlark.String(jsonData), nil
 }
 
+// Implementation of the `proto.to_any()` built-in function. Returns a
+// skyProtoMessage with an `Any` proto.Message in it.
+func fnProtoToAny(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var msg *skyProtoMessage
+	err := wantSingleProtoMessage("proto.to_any", args, kwargs, &msg)
```
dilyevsky:
```suggestion
	err := wantSingleProtoMessage(fn.Name(), args, kwargs, &msg)
```

idiamond-stripe:
Hit this today when trying to use the Envoy API in skycfg. The Envoy `Filter` message takes a `google.protobuf.Any`, which isn't currently possible to implement using Skycfg [0]. This typed config is one of the options in the `extensions/` subdirectory, for example `HTTPConnectionManager` [1].

It's not currently possible to work around as it's not possible to marshal Skycfg to a protobuf so you can put it into the `value` block of `google.protobuf.Any`, nor introspect the type name.

[0] https://github.com/envoyproxy/data-plane-api/blob/master/envoy/api/v2/listener/listener_components.proto#L29-L43
[1] https://github.com/envoyproxy/data-plane-api/blob/master/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#L41

jmillikin-stripe:
Replaced by https://github.com/stripe/skycfg/pull/83

PR #47, opened 2019-03-23
[yaml]: support yaml.unmarshal

charleschengxu:


internal/go/skycfg/yaml.go
```diff
@@ -60,3 +64,61 @@ func fnYamlMarshal(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple
...
+func yamlUnmarshal() starlark.Callable {
+	return starlark.NewBuiltin("yaml.unmarshal", fnYamlUnmarshal)
+}
+
+func fnYamlUnmarshal(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var filename string
+	if err := starlark.UnpackPositionalArgs(fn.Name(), args, nil, 1, &filename); err != nil {
+		return nil, err
+	}
+	blob, err := ioutil.ReadFile(filename)
```
jmillikin-stripe: Reading in a file will make `yaml.unmarshal()` non-hermetic. Can you change it to accept the YAML content as its parameter? That way the core Skycfg library remains safe when processing untrusted inputs.

In your own code you could add `read_file(filename) -> content`, and use it like:

```python
yaml.unmarshal(read_file("filename.yaml"))
```

charleschengxu: I agree that accepting yaml content makes more sense. I might have missed it in this repo but it seems `read_file` has yet to be implemented, so in the test cases I used `yaml.marshal()` instead.

When I get some time, I can come back to do `read_file` as another builtin if you do not mind.

jmillikin-stripe: Ah, sorry, I meant that `read_file()` should be defined in your own code. Skycfg doesn't have any functions that can inspect the filesystem, by design.

jmillikin-stripe:
Thank you for your contribution! I've merged this into the main branch.

PR #49, opened 2019-04-08
Handle type aliasing for string pointer fields.

dilyevsky:
Useful for some Kubernetes types where field is pointer to string alias.

Example (https://github.com/kubernetes/api/blob/8fc60343b75c9ba9a2b9c7cb5fc3b9c5e9b90158/admissionregistration/v1beta1/types.go#L242):

```
wh = proto.package("k8s.io.api.admissionregistration.v1beta1").Webhook(
    sideEffects = "Some",
)
...
```

(prior to this change would produce crash).


dilyevsky:
```
go build k8s.io/apimachinery/pkg/api/resource: module requires Go 1.12
```
whaaa?

dilyevsky:
Looks like `js.NewCallback` was deprecated. @jmillikin-stripe PTL

dilyevsky:
Hm I’ll have to somehow fake/mock protobuf msg with all its bells and
whistles (descriptors etc). I’ll see if there’s a good way of doing that

jmillikin-stripe:
I would take a hello world message .pb.go, copy-paste it into the test directory, and edit it to look like a Kubernetes-style message source.

dilyevsky:
Success! It actually needed quite little - main thing was gzipped FileDescriptorProto. @jmillikin-stripe PTAL

dilyevsky:
Thanks!

PR #53, opened 2019-06-25
starlark: use the new CallStack API

jbdalido:
related to https://github.com/google/starlark-go/pull/193/commits/efe33da344e187de1ef52c968c1ad19d5b6af259

Replacing calls to Caller by CallFrame and CallStack.


go.sum
```diff
@@ -1,13 +1,22 @@
...
 github.com/gogo/protobuf v1.2.0/go.mod h1:r8qH/GZQm5c6nD/R0oafs1akxWv10x8SbQlK7atdtwQ=
+github.com/gogo/protobuf v1.2.1 h1:/s5zKNz0uPFCZ5hddgPdo2TK2TVrUNMn0OOX8/aZMTE=
+github.com/gogo/protobuf v1.2.1/go.mod h1:hp+jE20tsWTFYpLwKvXlhS1hjn+gTNwPg2I6zVXpSg4=
 github.com/golang/protobuf v1.2.0 h1:P3YflyNX/ehuJFLhxviNdFxQPkGK5cDcApsge1SqnvM=
 github.com/golang/protobuf v1.2.0/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
+github.com/golang/protobuf v1.3.1 h1:YF8+flBXS5eO826T4nzqPrxfhQThhXl0YzfuUPu4SBg=
+github.com/golang/protobuf v1.3.1/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
 github.com/jmillikin-stripe/godebug v0.0.0-20180620173319-8279e1966bc1 h1:JgsVrDAUy59N248f3l4RGZ0hij5u1HTit8iJr1mFSBY=
 github.com/jmillikin-stripe/godebug v0.0.0-20180620173319-8279e1966bc1/go.mod h1:gFqr/IKD8P+Hluq9gThCR944BAu6jUqd5H/R3PrPfuM=
+github.com/kisielk/errcheck v1.1.0/go.mod h1:EZBBE59ingxPouuu3KfxchcWSUPOHkagtvWXihfKN4Q=
```
jmillikin-stripe: Where are these new dependencies coming from? Are they a side-effect from one of the dependency upgrades?

skycfg.go
```diff
@@ -69,7 +69,7 @@ func (r *localFileReader) Resolve(ctx context.Context, name, fromPath string) (s
 	if filepath.Separator != '/' && strings.ContainsRune(name, filepath.Separator) {
 		return "", fmt.Errorf("load(%q): invalid character in module name", name)
 	}
-	resolved := filepath.Join(r.root, filepath.FromSlash(path.Clean("/"+name)))
+	resolved := filepath.FromSlash(path.Clean(name))
```
jmillikin-stripe: What is this change for? It's removing `r.root` from the expression.

skycfg.go
```diff
@@ -223,8 +223,8 @@ func loadImpl(ctx context.Context, opts *loadOptions, filename string) (starlark
 	var load func(thread *starlark.Thread, moduleName string) (starlark.StringDict, error)
 	load = func(thread *starlark.Thread, moduleName string) (starlark.StringDict, error) {
 		var fromPath string
-		if thread.TopFrame() != nil {
-			fromPath = thread.TopFrame().Position().Filename()
+		if thread.CallStack() != nil {
+			fromPath = thread.CallStack().String()
```
jmillikin-stripe: `fromPath` is a file path, which is not what I would expect to get from `thread.CallStack().String()`.

internal/go/skycfg/assert.go
```diff
@@ -137,18 +138,20 @@ func (t *TestContext) AssertBinaryImpl(op syntax.Token) func(thread *starlark.Th
 		passes, err := starlark.Compare(op, val1, val2)
 		if err != nil {
 			buf := new(strings.Builder)
-			thread.Caller().WriteBacktrace(buf)
+			stk := thread.CallFrame(1)
+			fmt.Fprintf(buf, "%s", stk)
```
jmillikin-stripe: It looks like this `buf` isn't being used, so changing the error conditional to `if err != nil { return nil, err }` would clean it up.

internal/go/skycfg/assert.go
```diff
@@ -113,9 +113,10 @@ func (t *TestContext) CallInternal(thread *starlark.Thread, args starlark.Tuple,
 
 	if !val {
 		buf := new(strings.Builder)
-		thread.Caller().WriteBacktrace(buf)
+		stk := thread.CallFrame(1)
+		fmt.Fprintf(buf, "%s", stk)
 		err := assertionError{
-			position:  thread.Caller().Position().String(),
+			position:  stk.Pos.String(),
```
jmillikin-stripe: Rather than duplicating the stack trace formatting logic, the frame could be stored in `assertionError` and the formatting handled in `assertionError.Error()`. Something like this:
```go
err := assertionError{
	stack: thread.CallStack(),
}
```

internal/go/skycfg/assert.go
```diff
@@ -113,9 +113,10 @@ func (t *TestContext) CallInternal(thread *starlark.Thread, args starlark.Tuple,
 
 	if !val {
 		buf := new(strings.Builder)
-		thread.Caller().WriteBacktrace(buf)
+		stk := thread.CallFrame(1)
+		fmt.Fprintf(buf, "%s", stk)
```
jmillikin-stripe: I'm confused about the content of `buf` at this point. The docs and definition for `CallFrame` imply it's a single frame from the stack, so stringifying it won't provide a full stack trace.

jmillikin-stripe:
Timing out -- I've merged https://github.com/stripe/skycfg/pull/54 to update the starlark-go dependency. If the other changes in this PR (e.g. path resolution) are something you still want, feel free to re-open or re-branch.

PR #56, opened 2019-07-23
autoboxing scalars to their corresponding wrapper types

yq-stripe:
If a message field is typed as a wrapper type (e.g. StringValue, BoolValue, IntValue) don't throw a type error, but autobox them correspondingly. See below, 

```
syntax = "proto3";
package foo;
import "google/protobuf/wrappers.proto";

message packet { google.protobuf.StringValue val = 1; }
```

```
wkt = proto.package("google.protobuf")
foo = proto.package("foo")

def main(ctx):
    return [foo.packet(val = wkt.StringValue(value = "something"))]
    // instead we can write foo.packet(val = "something")
```

Thoughts?

### DIFF
--- a/internal/go/skycfg/proto_message.go
+++ b/internal/go/skycfg/proto_message.go
@@ -28,6 +28,7 @@ import (
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/wrappers"
 	dpb "github.com/golang/protobuf/ptypes/duration"
 	"go.starlark.net/starlark"
 	"go.starlark.net/syntax"
@@ -358,6 +359,29 @@ func maybeConvertString(v reflect.Value, t reflect.Type) (reflect.Value, bool) {
 func valueFromStarlark(t reflect.Type, sky starlark.Value) (reflect.Value, error) {
 	switch sky := sky.(type) {
 	case starlark.Int, starlark.Float, starlark.String, starlark.Bool:
+		switch t {
+			// autobox if t is a wrapper type
+			case reflect.TypeOf(&wrappers.StringValue{}):
+				if strval, ok := sky.(starlark.String); ok {
+					val := reflect.New(reflect.TypeOf(&wrappers.StringValue{}).Elem())
+					val.Elem().Set(reflect.ValueOf(wrappers.StringValue{Value: string(strval)}))
+					return val, nil
+				}
+			case reflect.TypeOf(&wrappers.Int32Value{}):
+				if intval, ok := sky.(starlark.Int); ok {
+					val := reflect.New(reflect.TypeOf(&wrappers.Int32Value{}).Elem())
+					int64val, _ := intval.Int64()
+					val.Elem().Set(reflect.ValueOf(wrappers.Int32Value{Value: int32(int64val)}))
+					return val, nil
+				}
+			case reflect.TypeOf(&wrappers.BoolValue{}):
+				if boolval, ok := sky.(starlark.Bool); ok {
+					val := reflect.New(reflect.TypeOf(&wrappers.BoolValue{}).Elem())
+					val.Elem().Set(reflect.ValueOf(wrappers.BoolValue{Value: bool(boolval)}))
+					return val, nil
+				}
+		}
+
 		scalar, err := scalarFromStarlark(t, sky)
 		if err != nil {
 			return reflect.Value{}, err
--- a/internal/go/skycfg/proto_message.go
+++ b/internal/go/skycfg/proto_message.go
@@ -368,52 +368,90 @@ func maybeConvertToWrapper(t reflect.Type, sky starlark.Value) (*reflect.Value, 
 	Int32ValueType := reflect.TypeOf(&wrappers.Int32Value{})
 	Int64ValueType := reflect.TypeOf(&wrappers.Int64Value{})
 	BytesValueType := reflect.TypeOf(&wrappers.BytesValue{})
+	UInt32ValueType := reflect.TypeOf(&wrappers.UInt32Value{})
+	UInt64ValueType := reflect.TypeOf(&wrappers.UInt64Value{})
 
-	switch sky := sky.(type) {
-	case starlark.Bool:
-		if t == BoolValueType {
-			val := reflect.New(t.Elem())
-			val.Elem().Set(reflect.ValueOf(wrappers.BoolValue{Value: bool(sky)}))
-			return &val, nil
-		}
-	case starlark.Int:
-		switch t {
-		case Int64ValueType:
-			val := reflect.New(t.Elem())
-			int64val, _ := sky.Int64()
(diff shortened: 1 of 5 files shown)


yq-stripe:
@jmillikin-stripe updated the PR with the suggested changes and added more wrapper types.

internal/go/skycfg/proto_message.go
```diff
@@ -355,9 +356,79 @@ func maybeConvertString(v reflect.Value, t reflect.Type) (reflect.Value, bool) {
...
+	StringValueType := reflect.TypeOf(&wrappers.StringValue{})
+	DoubleValueType := reflect.TypeOf(&wrappers.DoubleValue{})
+	Int32ValueType := reflect.TypeOf(&wrappers.Int32Value{})
+	Int64ValueType := reflect.TypeOf(&wrappers.Int64Value{})
+	BytesValueType := reflect.TypeOf(&wrappers.BytesValue{})
+
+	switch sky := sky.(type) {
+	case starlark.Bool:
+		if t == BoolValueType {
+			val := reflect.New(t.Elem())
```
jmillikin-stripe: If I remember correctly, the idiomatic way to do this is:
```go
val := reflect.ValueOf(&wrappers.BoolValue{Value: bool(sky)})
return &val, nil
```

internal/go/skycfg/proto_message.go
```diff
@@ -355,9 +356,79 @@ func maybeConvertString(v reflect.Value, t reflect.Type) (reflect.Value, bool) {
...
+			val.Elem().Set(reflect.ValueOf(wrappers.BoolValue{Value: bool(sky)}))
+			return &val, nil
+		}
+	case starlark.Int:
+		switch t {
+		case Int64ValueType:
+			val := reflect.New(t.Elem())
+			int64val, _ := sky.Int64()
```
jmillikin-stripe: You'll want to check the `ok` component of the return values, because Starlark integers can be arbitrarily large. Returning `nil, nil` in that situation should be fine because it'll fall through to a generic assignment error.

internal/go/skycfg/proto_message.go
```diff
@@ -355,9 +356,79 @@ func maybeConvertString(v reflect.Value, t reflect.Type) (reflect.Value, bool) {
...
+// protobuf wrapper type and attempts to convert it.
+// Returns a non-nil Value and nil error if successfully converted, a non-nil error
+// if unsuccessful conversion, and nil Value and nil error if no conversion attempt
+// was required.
+func maybeConvertToWrapper(t reflect.Type, sky starlark.Value) (*reflect.Value, error) {
+	BoolValueType := reflect.TypeOf(&wrappers.BoolValue{})
+	StringValueType := reflect.TypeOf(&wrappers.StringValue{})
+	DoubleValueType := reflect.TypeOf(&wrappers.DoubleValue{})
+	Int32ValueType := reflect.TypeOf(&wrappers.Int32Value{})
+	Int64ValueType := reflect.TypeOf(&wrappers.Int64Value{})
```
jmillikin-stripe: We'll want to handle `UInt64Value` and `UInt32Value` too.

internal/go/skycfg/proto_message.go
```diff
@@ -355,9 +356,79 @@ func maybeConvertString(v reflect.Value, t reflect.Type) (reflect.Value, bool) {
...
+			return &val, nil
+		case DoubleValueType:
+			val := reflect.New(t.Elem())
+			int64val, _ := sky.Int64()
+			val.Elem().Set(reflect.ValueOf(wrappers.DoubleValue{Value: float64(int64val)}))
+			return &val, nil
+		case Int32ValueType:
+			val := reflect.New(t.Elem())
+			int64val, _ := sky.Int64()
+			if int64val >= math.MinInt32 && int64val <= math.MaxInt32 {
```
jmillikin-stripe: I believe you can use `starlark.AsInt32(sky)` to handle this error case.

internal/go/skycfg/proto_message.go
```diff
@@ -355,9 +356,79 @@ func maybeConvertString(v reflect.Value, t reflect.Type) (reflect.Value, bool) {
...
+	case starlark.Int:
+		switch t {
+		case Int64ValueType:
+			val := reflect.New(t.Elem())
+			int64val, _ := sky.Int64()
+			val.Elem().Set(reflect.ValueOf(wrappers.Int64Value{Value: int64val}))
+			return &val, nil
+		case DoubleValueType:
+			val := reflect.New(t.Elem())
+			int64val, _ := sky.Int64()
```
jmillikin-stripe: Use `starlark.AsFloat(sky)` to check for overflow.

internal/go/skycfg/proto_message.go
```diff
@@ -355,9 +356,79 @@ func maybeConvertString(v reflect.Value, t reflect.Type) (reflect.Value, bool) {
 	return v, true
 }
 
+// maybeConvertToWrapper checks if [sky] is a primitive and [t] is a corresponding
+// protobuf wrapper type and attempts to convert it.
+// Returns a non-nil Value and nil error if successfully converted, a non-nil error
+// if unsuccessful conversion, and nil Value and nil error if no conversion attempt
+// was required.
+func maybeConvertToWrapper(t reflect.Type, sky starlark.Value) (*reflect.Value, error) {
```
jmillikin-stripe: It might be worth reversing which type the switch is checking. If you `switch t`, then you can use the Starlark library's built-in conversion functions to handle cases like int->double.

internal/go/skycfg/proto_message.go
```diff
@@ -355,9 +356,79 @@ func maybeConvertString(v reflect.Value, t reflect.Type) (reflect.Value, bool) {
...
+	return nil, nil
+}
+
 func valueFromStarlark(t reflect.Type, sky starlark.Value) (reflect.Value, error) {
 	switch sky := sky.(type) {
 	case starlark.Int, starlark.Float, starlark.String, starlark.Bool:
+		value, err := maybeConvertToWrapper(t, sky)
+		if err != nil {
+			return reflect.Value{}, err
+		} else if value != nil && err == nil {
```
jmillikin-stripe: `} else if value != nil {` , we know `err == nil` always holds here because it's in an else branch.

internal/go/skycfg/proto_message.go
yq-stripe: Makes sense, though I think an overflow/value error description makes more sense here than a TypeError in the !ok case instead of falling through

internal/go/skycfg/proto_message.go
yq-stripe: yep!

internal/go/skycfg/skycfg_test.go
```diff
@@ -199,7 +297,7 @@ func TestSkycfgEndToEnd(t *testing.T) {
 			if err == nil {
 				t.Error(
 					"Bad err result from ExecMain for case", testCase.caseName,
-					"\nExpected nil",
```
yq-stripe: typo

PR #57, opened 2019-07-31
extend supported types for yaml

canthefason:
Currently, supported fields for yaml.unmarshal is limited with strings, maps and slices. This PR will extend it and will add int, uint, float, bool, null kinds to the supported fields. 

Signed-off-by: Jon Yucel <jon.yucel@getcruise.com>

### DIFF
--- a/internal/go/skycfg/yaml.go
+++ b/internal/go/skycfg/yaml.go
@@ -86,8 +86,19 @@ func fnYamlUnmarshal(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tup
 
 // toStarlarkValue is a DFS walk to translate the DAG from go to starlark
 func toStarlarkValue(obj interface{}) (starlark.Value, error) {
+	if obj == nil {
+		return starlark.None, nil
+	}
 	rt := reflect.TypeOf(obj)
 	switch rt.Kind() {
+	case reflect.Int:
+		return starlark.MakeInt(obj.(int)), nil
+	case reflect.Uint64:
+		return starlark.MakeUint64(obj.(uint64)), nil
+	case reflect.Bool:
+		return starlark.Bool(obj.(bool)), nil
+	case reflect.Float64:
+		return starlark.Float(obj.(float64)), nil
 	case reflect.String:
 		return starlark.String(obj.(string)), nil
 	case reflect.Map:
@@ -114,6 +125,6 @@ func toStarlarkValue(obj interface{}) (starlark.Value, error) {
 		}
 		return starlark.NewList(starvals), nil
 	default:
-		return nil, fmt.Errorf("%v is not a slice, map, or string", obj)
+		return nil, fmt.Errorf("%s (%v) is not a supported type", rt.Kind(), obj)
 	}
 }
--- a/internal/go/skycfg/yaml.go
+++ b/internal/go/skycfg/yaml.go
@@ -90,17 +90,18 @@ func toStarlarkValue(obj interface{}) (starlark.Value, error) {
 		return starlark.None, nil
 	}
 	rt := reflect.TypeOf(obj)
+	v := reflect.ValueOf(obj)
 	switch rt.Kind() {
-	case reflect.Int:
-		return starlark.MakeInt(obj.(int)), nil
-	case reflect.Uint64:
-		return starlark.MakeUint64(obj.(uint64)), nil
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		return starlark.MakeInt64(v.Int()), nil
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		return starlark.MakeUint64(v.Uint()), nil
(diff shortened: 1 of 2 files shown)


internal/go/skycfg/yaml.go
```diff
@@ -86,8 +86,19 @@ func fnYamlUnmarshal(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tup
 
 // toStarlarkValue is a DFS walk to translate the DAG from go to starlark
 func toStarlarkValue(obj interface{}) (starlark.Value, error) {
+	if obj == nil {
+		return starlark.None, nil
+	}
 	rt := reflect.TypeOf(obj)
 	switch rt.Kind() {
+	case reflect.Int:
```
jmillikin-stripe: We also need to handle `reflect.Int64`, which `yaml.Unmarshal()` will use for integers in the range `(math.MaxInt32, math.MaxInt64]`: https://play.golang.org/p/aPb8Ml_cYSB

I'm also a little nervous about being so precise about which integer types the YAML parser returns. Can this code instead handle all five signed integer kinds in the same `case` and pass to `starlark.MakeInt64()`?

internal/go/skycfg/yaml.go
```diff
@@ -86,8 +86,19 @@ func fnYamlUnmarshal(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tup
...
+		return starlark.MakeInt(obj.(int)), nil
+	case reflect.Uint64:
```
jmillikin-stripe: Similarly, lets handle all five unsigned integer kinds here.

internal/go/skycfg/yaml.go
```diff
@@ -86,8 +86,19 @@ func fnYamlUnmarshal(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tup
...
+		return starlark.MakeUint64(obj.(uint64)), nil
+	case reflect.Bool:
+		return starlark.Bool(obj.(bool)), nil
+	case reflect.Float64:
```
jmillikin-stripe: and `Float32`

internal/go/skycfg/yaml_test.go
```diff
@@ -91,7 +91,7 @@ func TestYamlToSky(t *testing.T) {
 	env := starlark.StringDict{
 		"yaml": YamlModule(),
 	}
-	skyExpr := `{"strKey": "val", "arrKey": ["a", "b"], "mapKey": {"subkey": "val"}}`
+	skyExpr := `{"strKey": "val", "arrKey": ["a", "b"], "mapKey": {"subkey": "val"}, "uintKey": 12345678901234567890, "intKey": 123, "floatKey": 1.234, "boolKey": False, "nullKey": None}`
```
jmillikin-stripe: This line is a bit hard to read now that it's scrolling past the edge of Github's diff column. Can you add newlines?

internal/go/skycfg/yaml_test.go
```diff
@@ -121,6 +121,31 @@ func TestYamlToSky(t *testing.T) {
 			key:  "mapKey",
 			want: `{"subkey": "val"}`,
 		},
+		{
+			name: "key mapped to Uint",
+			key:  "uintKey",
+			want: `12345678901234567890`,
```
jmillikin-stripe: I recommend testing specific special boundary values such as `math.MaxInt32` ± 1. It might also be good to have a test for parsing `math.MaxInt64+1`, which will hopefully fail with some reasonable error message.

canthefason:
@jmillikin-stripe PTAL

internal/go/skycfg/yaml.go
```diff
@@ -86,10 +86,22 @@ func fnYamlUnmarshal(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tup
...
 	rt := reflect.TypeOf(obj)
+	v := reflect.ValueOf(obj)
 	switch rt.Kind() {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		return starlark.MakeInt64(v.Int()), nil
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		return starlark.MakeUint64(v.Uint()), nil
+	case reflect.Bool:
+		return starlark.Bool(v.Bool()), nil
+	case reflect.Float32, reflect.Float64:
```
dilyevsky: Since skycfg has floats disabled, probably shouldn't be supporting these on the ingest either.

jmillikin-stripe: The `skycfg` package doesn't have a position on whether floats are enabled or not. This is controlled by a global mutable variable in the `starlark` package.

dilyevsky: Fair enough. Should this test for `resolve.AllowFloat` though so that floats don't accidentally leak into configs even though they are disabled?

jmillikin-stripe: That sounds reasonable, but it would probably need to be at a lower level since floats can also arrive via JSON or Protobuf parsing.

PR #58, opened 2019-07-31
add support for more yaml key/value types

yq-stripe:
Right now module yaml's unmarshal only supports string key types, and slice, string, or map values. This seems limited. `yaml.marshal` handles translating ints and bool values completely fine so why not the same for `yaml.unmarshal`? 

- add support for key types: ints, null (~ in yaml) 
- add support for value types: bool, ints, null 
- TODO add yamlToSky tests

@jmillikin-stripe


dilyevsky:
This is duplicate of #57

yq-stripe:
Didn't notice that - glad to see it's already happening! closing in favor of #57 then

Issue #64, opened 2019-09-30
Go 1.13's `go mod tidy` fails to import skycfg generated test protos


jbarrick-mesosphere:
Currently the test_proto package is not checked in to the repository, which makes `go mod tidy` fail:

```
$ go mod tidy -v    
go: finding github.com/stripe/skycfg latest
github.com/mesosphere/pumpkin/pkg/parser/starlark imports
	github.com/stripe/skycfg imports
	github.com/stripe/skycfg/internal/go/skycfg tested by
	github.com/stripe/skycfg/internal/go/skycfg.test imports
	github.com/stripe/skycfg/test_proto: module github.com/stripe/skycfg@latest (v0.0.0-20190821211108-c7a2dc23be62) found, but does not contain package github.com/stripe/skycfg/test_proto
```

Is it possible to check the package in?

It seems that `go mod tidy` does import test dependencies: https://github.com/golang/go/issues/26626

jmillikin-stripe:
It looks like this was a change in `go mod` behavior introduced in Go 1.13. Checking in generated code is a bad habit to get into, so I'll see if a workaround is possible via a stub file.

jmillikin-stripe:
https://github.com/stripe/skycfg/issues/64 should fix this when merged.

jmillikin-stripe:
Stubs have been checked in, `go mod tidy` v1.13 should work for you now.

jbarrick-mesosphere:
Thanks! Works like a charm.

PR #65, opened 2019-10-03
Move testdata protos under internal/ and use more compatible paths.

jmillikin-stripe:
Fixes https://github.com/stripe/skycfg/issues/64


jmillikin-stripe:
R= @abloom-stripe | @mmoriarity-stripe | @kklipsch-stripe

PR #66, opened 2019-12-04
Add `assert.fails` to check if a function fails.

mmoriarity-stripe:
Example:

```python
def might_fail(input):
  if input < 10:
    fail("input is less than 10")
  else:
    return input

def test_might_fail(t):
  t.assert.fails(might_fail, 3)
  # passes

  t.assert.fails(might_fail, 14)
  # assertion failure: function might_fail should have failed
```

### DIFF
--- a/internal/go/skycfg/assert.go
+++ b/internal/go/skycfg/assert.go
@@ -87,9 +87,6 @@ func (err assertionError) Error() string {
 type TestContext struct {
 	Attrs    starlark.StringDict
 	Failures []error
-
-	// error returned from the `fail` builtin, for `assert.fails`
-	FailError error
 }
 
 var _ starlark.HasAttrs = (*Module)(nil)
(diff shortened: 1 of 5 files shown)


internal/go/skycfg/assert.go
```diff
@@ -153,6 +165,32 @@ func (t *TestContext) AssertBinaryImpl(op syntax.Token) func(thread *starlark.Th
...
+func (t *TestContext) AssertFails(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	// reset this state so we can tell if this test triggers a fail
+	t.FailError = nil
+
+	failFn := args[0]
+	failArgs := args[1:]
+	_, err := starlark.Call(thread, failFn, failArgs, kwargs)
+	if err != nil {
+		// if the result was an error from `fail()`, the assertion passes
+		// starlark swallows the error type in Call, so instead we track it on the test context
```
jmillikin-stripe: I'm not sure what this is verifying -- is it trying to distinguish calls to `fail()` differently from other sources of failure? I'm not sure we want to be that granular, since calls like `fail('Error: key "abc" not in dict')` are indistinguishable to the user from "built-in" errors that might come from the Starlark runtime.

In addition, we'll want this to return some value representing the error so that the user can inspect that it's the correct error for the expected condition. A struct with a `string message` field would probably be fine, since Starlark doesn't have the concept of an error hierarchy.

mmoriarity-stripe: This makes a lot of sense: for some reason I had it in my head that only errors from `fail` should count, so I initially tried to make them a distinct type (`FailError`) but found that Starlark doesn't preserve those types, everything comes out an `EvalError`. So I tracked it in the test context. But you're right, any error that came up here should count, and that simplifies the implementation quite a bit.

And we can definitely have `assert.fails` return a small struct with the error message, presumably for a subsequent assert on the message.

PR #69, opened 2020-01-03
Generated empty messages without `proto.Clone`.

### LINKED ISSUE #27: whats the correct way to convert non protobuf fields ?
First thank you so much for open sourcing this tool !

```
  volumeClaimTemplates:
  - metadata:
      name: postgres-data
    spec:
      accessModes: [ "ReadWriteOnce" ]
      resources:
        requests:
          storage: 2Gi
```

I am trying to convert the above from yaml to skycfg. What would be the correct way of defining `spec.resources.requests` ?

chhsia0:
This PR partially fixes #27.

The `resource.Quantity` proto message is generated from a go struct through the `go-to-protobuf` k8s tool. This creates two problems:
* The `Quantity` go struct has unexported fields, which is not cloneable through reflection, but `skycfg` uses `proto.Clone` to create empty messages. This problem is fixed in this PR by creating empty messages through `reflect.New` instead.
* The Quantity go struct doesn’t have any protobuf tag, but `skycfg` uses tags to extract field names. As a result, skycfg doesn’t not recognize `resource.Quantity(string = "1000m")`. To resolve this, you'll need to replace `k8s.io/apimachinery` in `go.mod` with my fork, which adds protobuf tags to the `Quantity` go struct:
```
replace k8s.io/apimachinery => github.com/chhsia0/apimachinery <k8s_version>-quantity-tag
```
where k8s_version can be 1.14, 1.15, 1.16 or 1.17. I'll post a PR to upstream this change.

### DIFF
--- a/internal/go/skycfg/proto_api.go
+++ b/internal/go/skycfg/proto_api.go
@@ -270,7 +270,7 @@ func fnProtoFromText(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tup
 	if !ok {
 		return nil, fmt.Errorf("%s: for parameter 2: got %s, want proto.MessageType", "proto.from_text", msgType.Type())
 	}
-	msg := proto.Clone(protoMsgType.emptyMsg)
+	msg := protoMsgType.emptyMsg()
 	msg.Reset()
 	if err := proto.UnmarshalText(string(value), msg); err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


jmillikin-stripe:
Please file a bug against Kubernetes, asking them to use valid Protobuf implementations instead of doing their own half-baked thing. Skycfg is designed to operate on Protobuf messages, not arbitrary Go structs, and we don't have plans to support Go structs that falsely claim to be Protobuf messages.

Note that the Kubernetes code may break anyway once https://github.com/golang/protobuf/issues/364 lands in upstream go-protobuf. And it will _definitely_ not work once support for dynamic messages is implemented in go-protobuf, since at that point a message type might not be backed by any concrete Go type at all.

You may be able to work around this by updating your build workflow to generate Go source from the Kubernetes .proto files using the standard Protobuf tooling. This is what we (Stripe) do, using Bazel as our build tool. Other build tools -- Make, Rake, CMake, basically anything that can run commands -- should also be usable in this role.

dilyevsky:
fwiw, we wrapped this thing with a special helper in Isopod (https://github.com/cruise-automation/isopod/blob/19a7f547c270318529df74a59f47bc5684837ec8/pkg/kube/util.go#L29) and never had any issues.

jbarrick-mesosphere:
Thank you @dilyevsky !

PR #72, opened 2020-01-28
[yaml]: Update YAML parser to yaml.v3

dilyevsky:
Migrate to https://github.com/go-yaml/yaml/tree/v3 and don't roundtrip through JSON to support stable order of YAML maps (perviously they would be shuffled due to round-tripping via Go map).

This can also enable support for alias YAML nodes but is not part of this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/stripe/skycfg/72)

### DIFF
--- a/internal/go/skycfg/yaml.go
+++ b/internal/go/skycfg/yaml.go
@@ -20,9 +20,10 @@ import (
 	"bytes"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"go.starlark.net/starlark"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
 )
 
 // YamlModule returns a Starlark module for YAML helpers.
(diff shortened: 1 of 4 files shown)


internal/go/skycfg/yaml.go
```diff
@@ -44,24 +45,101 @@ func yamlMarshal() starlark.Callable {
...
+	enc := yaml.NewEncoder(&buf)
+	enc.SetIndent(2)
+	if err := enc.Encode(&yaml.Node{
+		Kind:    yaml.DocumentNode,
+		Content: []*yaml.Node{node},
+	}); err != nil {
 		return nil, err
 	}
-	return starlark.String(yamlBytes), nil
+	enc.Close()
```
jmillikin-stripe: Does the error from `Close()` need to be checked?

PR #73, opened 2020-02-03
Fix construction of multiply-nested message types.

jmillikin-stripe:
Given a Protobuf message layout like this:
```protobuf
message A {
  message B {
    message C {
      // ...
    }
  }
}
```

The code for calculating message names was incorrect; it would drop `A.` from the lookup path.

This PR fixes it by recording the package-relative qualified name as a field of `skyProtoMessageType`.

r? @abloom-stripe

### DIFF
--- a/internal/go/skycfg/proto_message_type.go
+++ b/internal/go/skycfg/proto_message_type.go
@@ -27,11 +27,11 @@ import (
 	"go.starlark.net/starlark"
 )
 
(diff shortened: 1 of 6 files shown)


jmillikin-stripe:
Note: failed build is because `github.com/golang/protobuf` has dropped support for Go v1.11, and our CI script isn't pinning to an older version.

PR #74, opened 2020-02-27
Add a WithEntryPoint ExecOption

abloom-stripe:
Allow overriding the `main` entry point in your skycfg code with another named function as long as its signature is the same.

```python
# skycfg
def not_main(ctx):
    message = test_proto.MessageV2()
    return [message]
```

```go
// go
config.Main(context.Background(), skycfg.WithVars(testCase.vars), skycfg.WithEntryPoint("not_main"))
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,3 +9,5 @@ require (
 )
 
 replace github.com/kylelemons/godebug => github.com/jmillikin-stripe/godebug v0.0.0-20180620173319-8279e1966bc1
+
+go 1.13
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -9,3 +9,5 @@ require (
 )
 
 replace github.com/kylelemons/godebug => github.com/jmillikin-stripe/godebug v0.0.0-20180620173319-8279e1966bc1
+
+go 1.13
```
abloom-stripe: My editor or local build process seems to have added this automatically but I'll be honest I don't know why or if it's necessary. Please let me know if I should remove it from this PR.

abloom-stripe:
rebased and removed my commit deleting go 1.11.

PTAL @jmillikin-stripe

Issue #76, opened 2020-05-20
Unexpected behaviour when checking equality of enums


tan-stripe:
Below is a code snippet showing the unexpected behaviour when comparing 2 different enum variables in skycfg
```
envoy = proto.package("envoy.api.v2")

def main(ctx):
  a = envoy.Cluster.LbPolicy.ROUND_ROBIN
  b = envoy.Cluster.LbPolicy.ROUND_ROBIN
  # this returns false
  print(a == b)
  # however, this works
  print(str(a) == str(b))
  return []
```

jmillikin-stripe:
Thanks for the bug report! This should be fixed at master HEAD.

PR #77, opened 2020-05-21
Implement starlark.Comparable for enums.

dilyevsky:
Implement `starlark.Comparable` interface for `skyProtoEnumValue` to handle enum comparison correctly. Only supports `==` and `!=` operations.

Fixes https://github.com/stripe/skycfg/issues/76

### DIFF
--- a/internal/go/skycfg/proto_enum.go
+++ b/internal/go/skycfg/proto_enum.go
@@ -23,6 +23,7 @@ import (
 
 	descriptor_pb "github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"go.starlark.net/starlark"
+	"go.starlark.net/syntax"
 )
 
 type skyProtoEnumType struct {
@@ -64,6 +65,8 @@ type skyProtoEnumValue struct {
 	value     int32
 }
 
+var _ starlark.Comparable = (*skyProtoEnumValue)(nil)
+
 func (v *skyProtoEnumValue) String() string {
 	return fmt.Sprintf("<%s %s=%d>", v.typeName, v.valueName, v.value)
 }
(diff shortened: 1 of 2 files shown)


jmillikin-stripe:
Hah, I was literally just working on this and saw the github notification after I created https://github.com/stripe/skycfg/pull/78.

I'll add some comments here too.

internal/go/skycfg/proto_enum.go
```diff
@@ -74,6 +77,24 @@ func (v *skyProtoEnumValue) Hash() (uint32, error) {
 	return starlark.MakeInt64(int64(v.value)).Hash()
 }
 
+func (v *skyProtoEnumValue) CompareSameType(op syntax.Token, y starlark.Value, depth int) (bool, error) {
+	other, ok := y.(*skyProtoEnumValue)
+	if !ok {
+		return false, fmt.Errorf("`%v' not enum value", y)
+	}
+
+	eq := (v.typeName == other.typeName) && (v.valueName == other.valueName) && (v.value == other.value)
```
jmillikin-stripe, line 86: This is going to return not-equal for enum values with the same integer code but different names, which is different from how `Hash()` works -- I believe that would cause problems for code that uses enum values as dict keys.

internal/go/skycfg/proto_enum.go
```diff
@@ -74,6 +77,24 @@ func (v *skyProtoEnumValue) Hash() (uint32, error) {
 	return starlark.MakeInt64(int64(v.value)).Hash()
 }
 
+func (v *skyProtoEnumValue) CompareSameType(op syntax.Token, y starlark.Value, depth int) (bool, error) {
+	other, ok := y.(*skyProtoEnumValue)
```
jmillikin-stripe, line 81: Using upstream Starlark as an example, it seems that `CompareSameType` is allowed to assume its input `Value` is already of the correct type (and `panic` if this requirement isn't met).

See https://github.com/google/starlark-go/blob/0aa95694c768/starlark/value.go#L714-L715:
```go
func (x *Dict) CompareSameType(op syntax.Token, y_ Value, depth int) (bool, error) {
	y := y_.(*Dict)
	switch op {
	case syntax.EQL:
```
```

dilyevsky: Interesting. Done

internal/go/skycfg/proto_enum.go
dilyevsky: Good point. Fixed

PR #82, opened 2020-09-28
Add Envoy Skycfg Example

idiamond-stripe:
This PR adds a very simple Skcyfg example that runs an Envoy xDS using https://github.com/envoyproxy/go-control-plane. The example loads the provided config on start, and reloads the config on `SIGHUP`.

 `go run main.go` launches the XDS running by default on `0.0.0.0:8080`. It loads the provided config, for example `go run envoy.cfg` and exposes an ADS interface and the default gRPC health server.

The `run-envoy.sh` script starts an instance of Envoy running in a container, mounting ports `9901`,`10000`, `10001`. There is a bootstrap config for Envoy which connects outside the container to an ADS instance on `8080` (described in YAML). There's a health check configured for the ADS to determine if it's gone away.

The ADS should inject the provided listeners and clusters creating two routes to example.com and example.org respectively on `:10000` and `:10001`. Visiting these at http://localhost:10000 or http://localhost:10001 produces a working example.com page for me: 

image


internal/go/skycfg/proto_api.go
```diff
@@ -45,16 +45,19 @@ func NewProtoModule(registry ProtoRegistry) *ProtoModule {
...
 			"from_json":    starlark.NewBuiltin("proto.from_json", fnProtoFromJson),
 			"from_text":    starlark.NewBuiltin("proto.from_text", fnProtoFromText),
 			"from_yaml":    starlark.NewBuiltin("proto.from_yaml", fnProtoFromYaml),
+			"from_proto":   starlark.NewBuiltin("proto.from_proto", fnProtoFromProto),
 			"merge":        starlark.NewBuiltin("proto.merge", fnProtoMerge),
 			"set_defaults": starlark.NewBuiltin("proto.set_defaults", fnProtoSetDefaults),
 			"to_json":      starlark.NewBuiltin("proto.to_json", fnProtoToJson),
 			"to_text":      starlark.NewBuiltin("proto.to_text", fnProtoToText),
 			"to_yaml":      starlark.NewBuiltin("proto.to_yaml", fnProtoToYaml),
+			"to_proto":     starlark.NewBuiltin("proto.to_proto", fnProtoToProto),
```
jmillikin-stripe, line 60: Let's add functions like `to_any`, which return the `google.protobuf.Any` directly. This way users don't have to deal with the internal details of the type URLs, and don't need to define their own marshal/unmarshal helpers.

internal/go/skycfg/proto_api.go
```diff
@@ -45,16 +45,19 @@ func NewProtoModule(registry ProtoRegistry) *ProtoModule {
 	mod := &ProtoModule{
 		Registry: registry,
 		attrs: starlark.StringDict{
+			"name":        starlark.NewBuiltin("proto.clone", fnProtoName),
```
jmillikin-stripe, line 48: I believe this functionality is already available via `type(msg)`.

go.mod
```diff
@@ -1,11 +1,22 @@
 module github.com/stripe/skycfg
 
+go 1.15
+
 require (
+	github.com/cncf/udpa/go v0.0.0-20200909154343-1f710aca26a9 // indirect
+	github.com/envoyproxy/go-control-plane v0.9.6 // indirect
```
jmillikin-stripe, line 7: It looks like something in your dev environment got confused about which `go.mod` to update. It also touched a bunch of files in the other examples. Can you revert these unrelated changes?

jmillikin-stripe (changes requested):
General comment: It might be better to separate the skycfg changes (such as `google.protobuf.Any` support) into a separate PR, since they are not Envoy-specific functionality.

For the Envoy parts: would it be possible to simplify the example? For example by converting it to a plain front-proxy of `http://example.com`

idiamond-stripe:
I've removed the `any` changes and landed them here: https://github.com/stripe/skycfg/pull/83.

I've also simplified the example to run on two ports (`10000` and `10001`) and proxied them to example.com and example.org respectively.

All the other changes from the first, in progress review here are minor. I've added a health check and improved some go modules by `go mod tidy`, but overall these are mainly cosmetic. I should probably document the way this example works, but can do so in a follow up PR.

jmillikin-stripe (approved):
LGTM

PR #83, opened 2020-09-29
Add a proto.to_any method

idiamond-stripe:
This PR uses the code from https://github.com/stripe/skycfg/pull/46 and responds to the review feedback on this PR.

I've removed the gogo code entirely to be consistent with the rest of the provided functions in `proto_api.go`, which don't support gogo protos.

This PR is also quite old so I've also rebased it onto master.

### DIFF
--- a/internal/go/skycfg/proto_api.go
+++ b/internal/go/skycfg/proto_api.go
@@ -27,7 +27,7 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
 	"go.starlark.net/starlark"
-	"gopkg.in/yaml.v2"
+	yaml "gopkg.in/yaml.v2"
 )
 
 // UNSTABLE extension point for configuring how protobuf messages are loaded.
@@ -247,16 +247,10 @@ func fnProtoToAny(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple,
 		return nil, err
 	}
 
-	// Disambiguate between golang encoded proto messages.
-	var any proto.Message
-	if "" != proto.MessageName(msg.msg) {
-		// Returns a golang any.Any type.
-		any, err = ptypes.MarshalAny(msg.msg)
-	} else {
-		return nil, fmt.Errorf("%s: could not get message name for %s", fn.Name(), msg.Type())
-	}
+	// Returns a golang any.Any type.
+	any, err := ptypes.MarshalAny(msg.msg)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%s: could not get message name for %s", fn.Name(), msg.Type())
 	}
 
 	return NewSkyProtoMessage(any), nil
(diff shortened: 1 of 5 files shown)


idiamond-stripe:
r? @jmillikin-stripe 

The build is failing but I'm confused as to why given the dependencies don't appear to have changed.

internal/go/skycfg/proto_api.go
```diff
@@ -25,8 +25,9 @@ import (
 
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/ptypes"
 	"go.starlark.net/starlark"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
```
jmillikin-stripe, line 30: Can you keep this import alias as-is? It's nice to have explicitly named imports when upstream uses a non-default package name.

internal/go/skycfg/proto_api.go
```diff
@@ -236,6 +238,30 @@ func fnProtoToJson(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple
...
+// Implementation of the `proto.to_any()` built-in function. Returns a
+// skyProtoMessage with an `Any` proto.Message in it.
+func fnProtoToAny(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var msg *skyProtoMessage
+	err := wantSingleProtoMessage(fn.Name(), args, kwargs, &msg)
+	if err != nil {
+		return nil, err
+	}
+
+	// Disambiguate between golang encoded proto messages.
```
jmillikin-stripe, line 250: I'm not sure why this branch is needed. Why not `any, err := ptypes.MarshalAny(msg.msg)`? Was there a case you encountered where a protobuf message has no type name?

jmillikin-stripe (changes requested):
Looks generally good -- two minor comments.

jmillikin-stripe:
> The build is failing but I'm confused as to why given the dependencies don't appear to have changed.

It looks like something in `Go: master` is unhappy with the contents of `go.mod`. No idea why, but since it's not affecting any stable release I think it's fine to ignore.

We should probably drop `Go: master` from the CI suite since I don't see an obvious way to make Travis test failures informational.

dilyevsky:
Likely remnant of the pr this is based on - it used this check to
differentiate between gogo and standard registries (which I’m not sure if
they are actually equivalent or not)

internal/go/skycfg/proto_api.go
idiamond-stripe: I think this is a holdover/cargo-cult from the first iteration of this change. I'll remove it.

internal/go/skycfg/proto_api.go
idiamond-stripe: I think this is a holdover/cargo-cult from the first iteration of this change. Let me change the alias back.

idiamond-stripe:
ptal @jmillikin-stripe

PR #84, opened 2020-11-14
Export proto package as a starlark value.

chhsia0:
This would allow users to create proto packages as globals in their own Starlark runtime.


jmillikin-stripe (approved):
LGTM

Thank you for the contribution!

PR #87, opened 2021-08-12
Move assert.go, fail.go to separate assertmodule

seena-stripe:
This is a pure refactor with no behavior change to separate the assertion/fail logic for skycfg tests from the legacy protobuf v1 implementation to make removing the package later cleaner.

I also
- Update go.sum to include dependencies (go.sum seemed stale, causing issues with VSCode go tooling)
- Add myself as a contributor


mmoriarity-stripe (approved):
LGTM

PR #88, opened 2021-08-13
Protobuf API V2: Add map, repeated, type conversions

seena-stripe:
## Summary

This PR adds a map and list implementation on top of the go protobuf v2 api, and adds type conversions (`protoreflect.Value` to/from `starlark.Value`). This is largely boilerplate, and is unused so it should not have any functional impact on Skycfg today.

- `protomodule_map` implementation to wrap an underlying `starlark.Dict` with typechecking on writes. This lines up with `protoMap` from the existing implementation https://github.com/stripe/skycfg/blob/trunk/internal/go/skycfg/proto_message.go#L869
- `protomodule_list` implementation to wrap an underlying `starlark.List` with typechecking on writes. This lines up with `protoRepeated` from the existing implementation https://github.com/stripe/skycfg/blob/trunk/internal/go/skycfg/proto_message.go#L679
- `type_conversions`: this implements `valueFromStarlark` and `valueToStarlark` like the existing implementation (throughout https://github.com/stripe/skycfg/blob/trunk/internal/go/skycfg/proto_message.go)

## Motivation
I am continuing the migration to the go protubuf v2 api, which is partially started on trunk

## Tests
- Added tests:
I have added tests for the `map` and `list` types. `type_conversions` will be tested implicitly once the message type is added (it's awkward to access right now). These tests cover `TestListMutation` and `TestMapMutation` from the existing implementation, but line up with the new `protomodule_test` functions for `messageType`, `enum` and `package`
- Ran bazel + go build
- I have another branch (https://github.com/stripe/skycfg/tree/seena/protobuf-api-v2) with the complete migration which passes all existing tests under `internal/`
- I have also done some snapshot tests on our internal usage of skycfg to verify the before/after behavior of the migration is consistent.

### DIFF
--- a/go/protomodule/protomodule_list.go
+++ b/go/protomodule/protomodule_list.go
@@ -31,7 +31,7 @@ var allowedListMethods = map[string]func(*protoRepeated) starlark.Value{
 }
 
 // protoRepeated wraps an underlying starlark.List to provide typechecking on
-// wrties
+// writes
 //
 // starlark.List is heterogeneous, where protoRepeated enforces all values
 // conform to the given fieldDesc
@@ -104,12 +104,7 @@ func (r *protoRepeated) Append(v starlark.Value) error {
 		return err
 	}
 
-	err = r.list.Append(v)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return r.list.Append(v)
 }
 
 func (r *protoRepeated) SetIndex(i int, v starlark.Value) error {
@@ -118,12 +113,7 @@ func (r *protoRepeated) SetIndex(i int, v starlark.Value) error {
 		return err
 	}
 
-	err = r.list.SetIndex(i, v)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return r.list.SetIndex(i, v)
 }
 
 func (r *protoRepeated) Extend(iterable starlark.Iterable) error {
--- a/go/protomodule/protomodule_list.go
+++ b/go/protomodule/protomodule_list.go
@@ -1,0 +1,191 @@
+// Copyright 2021 The Skycfg Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package protomodule
+
+import (
+	"fmt"
+
+	"go.starlark.net/starlark"
+	"go.starlark.net/syntax"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+var allowedListMethods = map[string]func(*protoRepeated) starlark.Value{
+	"clear":  nil,
+	"append": (*protoRepeated).wrapAppend,
+	"extend": (*protoRepeated).wrapExtend,
+}
+
+// protoRepeated wraps an underlying starlark.List to provide typechecking on
+// wrties
+//
+// starlark.List is heterogeneous, where protoRepeated enforces all values
+// conform to the given fieldDesc
+type protoRepeated struct {
+	fieldDesc protoreflect.FieldDescriptor
+	list      *starlark.List
+}
+
+var _ starlark.Value = (*protoRepeated)(nil)
+var _ starlark.Iterable = (*protoRepeated)(nil)
+var _ starlark.Sequence = (*protoRepeated)(nil)
+var _ starlark.Indexable = (*protoRepeated)(nil)
+var _ starlark.HasAttrs = (*protoRepeated)(nil)
+var _ starlark.HasSetIndex = (*protoRepeated)(nil)
+var _ starlark.HasBinary = (*protoRepeated)(nil)
+var _ starlark.Comparable = (*protoRepeated)(nil)
+
+func newProtoRepeated(fieldDesc protoreflect.FieldDescriptor) *protoRepeated {
+	return &protoRepeated{fieldDesc, starlark.NewList(nil)}
+}
+
+func newProtoRepeatedFromList(fieldDesc protoreflect.FieldDescriptor, l *starlark.List) (*protoRepeated, error) {
+	out := &protoRepeated{fieldDesc, l}
+	for i := 0; i < l.Len(); i++ {
+		err := scalarTypeCheck(fieldDesc, l.Index(i))
+		if err != nil {
+			return nil, err
+		}
+	}
+	return out, nil
+}
+
+func (r *protoRepeated) Attr(name string) (starlark.Value, error) {
+	wrapper, ok := allowedListMethods[name]
+	if !ok {
+		return nil, nil
+	}
+	if wrapper != nil {
+		return wrapper(r), nil
+	}
+	return r.list.Attr(name)
+}
+
+func (r *protoRepeated) AttrNames() []string                 { return r.list.AttrNames() }
+func (r *protoRepeated) Freeze()                             { r.list.Freeze() }
+func (r *protoRepeated) Hash() (uint32, error)               { return r.list.Hash() }
+func (r *protoRepeated) Index(i int) starlark.Value          { return r.list.Index(i) }
+func (r *protoRepeated) Iterate() starlark.Iterator          { return r.list.Iterate() }
+func (r *protoRepeated) Len() int                            { return r.list.Len() }
+func (r *protoRepeated) Slice(x, y, step int) starlark.Value { return r.list.Slice(x, y, step) }
+func (r *protoRepeated) String() string                      { return r.list.String() }
+func (r *protoRepeated) Truth() starlark.Bool                { return r.list.Truth() }
+
+func (r *protoRepeated) Type() string {
+	return fmt.Sprintf("list<%s>", typeName(r.fieldDesc))
+}
+
+func (r *protoRepeated) CompareSameType(op syntax.Token, y starlark.Value, depth int) (bool, error) {
+	other, ok := y.(*protoRepeated)
+	if !ok {
+		return false, nil
+	}
+
+	return starlark.CompareDepth(op, r.list, other.list, depth)
+}
+
+func (r *protoRepeated) Append(v starlark.Value) error {
+	err := scalarTypeCheck(r.fieldDesc, v)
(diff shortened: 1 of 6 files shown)


go/protomodule/protomodule_list.go
```diff
@@ -0,0 +1,191 @@
...
+			return nil, err
+		}
+	}
+	return out, nil
+}
+
+func (r *protoRepeated) Attr(name string) (starlark.Value, error) {
+	wrapper, ok := allowedListMethods[name]
+	if !ok {
+		return nil, nil
```
kathleen-stripe, line 70: OOC, is there a reason we don't want to return an error here? The intention/context here might be comment-worthy.

go/protomodule/protomodule_list.go
```diff
@@ -0,0 +1,191 @@
...
+	return starlark.CompareDepth(op, r.list, other.list, depth)
+}
+
+func (r *protoRepeated) Append(v starlark.Value) error {
+	err := scalarTypeCheck(r.fieldDesc, v)
+	if err != nil {
+		return err
+	}
+
+	err = r.list.Append(v)
```
kathleen-stripe: nit: I think you can just return `r.list.Append(v)` here.

go/protomodule/protomodule_list.go
```diff
@@ -0,0 +1,191 @@
...
+	return nil
+}
+
+func (r *protoRepeated) SetIndex(i int, v starlark.Value) error {
+	err := scalarTypeCheck(r.fieldDesc, v)
+	if err != nil {
+		return err
+	}
+
+	r.list.SetIndex(i, v)
```
kathleen-stripe: Looks like `err` wasn't reassigned to the return value of `r.list.SetIndex(i, v)` here, but same as above, we can just return `r.list.SetIndex(i, v)`.

go/protomodule/protomodule_list.go
```diff
@@ -0,0 +1,191 @@
...
+				return starlark.Binary(op, r.list, y)
+			case *protoRepeated:
+				return starlark.Binary(op, r.list, y.list)
+			}
+			return nil, nil
+		}
+		if side == starlark.Right {
+			if _, ok := y.(*starlark.List); ok {
+				return starlark.Binary(op, y, r.list)
+			}
```
kathleen-stripe, line 148: OOC, why is there no case for when `y` is a `protoRepeated` here?

go/protomodule/protomodule_map.go
```diff
@@ -0,0 +1,173 @@
...
+import (
+	"fmt"
+
+	"go.starlark.net/starlark"
+	"go.starlark.net/syntax"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+var allowedDictMethods = map[string]func(*protoMap) starlark.Value{
+	"clear":      nil,
```
kathleen-stripe, line 28: OOC, are there plans to support these functions in the future or is there a reason we're leaving them unimplemented?

go/protomodule/protomodule_map.go
```diff
@@ -0,0 +1,173 @@
...
+	return starlark.NewBuiltin("update", impl).BindReceiver(m)
+}
+
+func (m *protoMap) SetKey(k, v starlark.Value) error {
+	err := m.typeCheck(k, v)
+	if err != nil {
+		return err
+	}
+
+	err = m.dict.SetKey(k, v)
```
kathleen-stripe: Same here and on line 167 re: returning the return value of `m.dict.SetKey(k, v)` and `scalarTypeCheck(m.mapValue, v)` respectively.

go/protomodule/protomodule_map.go
```diff
@@ -0,0 +1,173 @@
...
+		if _, err := starlark.Call(thread, tempUpdate, args, kwargs); err != nil {
+			return nil, err
+		}
+		for _, item := range tempDict.Items() {
+			if err := m.SetKey(item[0], item[1]); err != nil {
+				return nil, err
+			}
+		}
+
+		return starlark.None, nil
```
kathleen-stripe, line 142: What is the return value for this function supposed to represent?

go/protomodule/protomodule_list.go
```diff
@@ -0,0 +1,191 @@
...
+func (r *protoRepeated) wrapAppend() starlark.Value {
+	impl := func(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+		var val starlark.Value
+		if err := starlark.UnpackPositionalArgs("append", args, kwargs, 1, &val); err != nil {
+			return nil, err
+		}
+		if err := r.Append(val); err != nil {
+			return nil, err
+		}
+		return starlark.None, nil
```
kathleen-stripe, line 164: OOC, what is the `starlark.Value` return value supposed to represent here?

go/protomodule/protomodule_list.go
```diff
@@ -0,0 +1,191 @@
...
+)
+
+var allowedListMethods = map[string]func(*protoRepeated) starlark.Value{
+	"clear":  nil,
+	"append": (*protoRepeated).wrapAppend,
+	"extend": (*protoRepeated).wrapExtend,
+}
+
+// protoRepeated wraps an underlying starlark.List to provide typechecking on
+// wrties
```
kathleen-stripe: nit: typo

go/protomodule/protomodule_test.go
```diff
@@ -200,6 +201,264 @@ func TestEnumType(t *testing.T) {
...
+    return l
+`,
+			want: `[]`,
+		},
+		{
+			exprFun: `
+def fun():
+    l = list()
+    l.extend(["a", "b"])
+    l[1] = "c"
```
kathleen-stripe, line 275: Does this syntax use `SetIndex` under the hood? If not, what do you think of adding explicit tests for `SetIndex`? Also, it looks like `Binary` isn't covered by tests.

go/protomodule/protomodule_test.go
```diff
@@ -200,6 +201,264 @@ func TestEnumType(t *testing.T) {
...
+    m["a"] = "some string"
+    m.clear()
+    return m
+`,
+			want: `{}`,
+		},
+		{
+			exprFun: `
+def fun():
+    l = list()
```
kathleen-stripe: Hmm....is this supposed to be a map?

go/protomodule/protomodule_test.go
```diff
@@ -200,6 +201,264 @@ func TestEnumType(t *testing.T) {
...
+			wantErr: errors.New(`TypeError: value 1 (type "int") can't be assigned to type "string".`),
+		},
+		{
+			expr:    `map().update([("a", 1)])`,
+			wantErr: errors.New(`TypeError: value 1 (type "int") can't be assigned to type "string".`),
+		},
+		{
+			expr:    `map().setdefault("a", 1)`,
+			wantErr: errors.New(`TypeError: value 1 (type "int") can't be assigned to type "string".`),
+		},
```
kathleen-stripe, line 441: Is it reasonable to add test cases for when the key is an int? (I assume that's also invalid and should error?)

go/protomodule/protomodule_test.go
```diff
@@ -200,6 +201,264 @@ func TestEnumType(t *testing.T) {
...
+    l = list()
+    l.extend(["a", "b"])
+    l[1] = 1
+    return l
+`,
+			wantErr: errors.New(`TypeError: value 1 (type "int") can't be assigned to type "string".`),
+		},
+	}
+	for _, test := range tests {
+		t.Run("", func(t *testing.T) {
```
kathleen-stripe: Could we give each of these tests names so that they're easier to debug if one of them fails? Same with the tests below.

mmoriarity-stripe (dismissed):

go/protomodule/type_conversions.go
```diff
@@ -0,0 +1,249 @@
...
+	_, err := scalarValueFromStarlark(fieldDesc, v)
+	return err
+}
+
+func typeError(fieldDesc protoreflect.FieldDescriptor, val starlark.Value, scalar bool) error {
+	expectedType := typeName(fieldDesc)
+
+	// FieldDescriptor has the same typeName for []string and string
+	// and typeError needs to distinguish setting a []string = int versus
+	// appending a value in []string
```
kathleen-stripe, line 224: Thanks for this helpful comment! OOC, why can't `typeName` handle this differentiation? i.e. return `[]string` for lists and `string` for string.

kathleen-stripe (changes requested):
A bunch of nits, but nothing major.
Feel free to send back to me for review if you disagree with the nits. Thanks!

go/protomodule/protomodule_map.go
seena-stripe: I copied the prior functionality to keep the change small but was also surprised to find `"clear": nil` means fallback to the underlying `starlark.List.Clear`

go/protomodule/protomodule_map.go
seena-stripe: The spec says `dict-update` returns `None` https://github.com/bazelbuild/starlark/blob/master/spec.md#dict%C2%B7update

go/protomodule/protomodule_list.go
seena-stripe: Same here, the spec just says it returns `None` https://github.com/bazelbuild/starlark/blob/master/spec.md#list%C2%B7append

go/protomodule/protomodule_test.go
seena-stripe: Yes exactly, good catch on the binary, going to add that as well

go/protomodule/protomodule_list.go
seena-stripe: Good question, I gave it a shot and I believe if you have `x + y` (both lists) it is calling `binary` on `x, left` falling to the other case, so we only have to support `[] + y` for starlark.List on the left side

go/protomodule/type_conversions.go
seena-stripe: I couldn't figure out a cleaner way to do this, but I'm bothered by it. The problem I was running into is that you have the cases:

``` python
msg.list_int_field = "a" # TypeError: value "a" (type "string") can't be assigned to "[]int"
msg.list_int_field = ["a"] # TypeError: value "a" (type "string") can't be assigned to "int"
```

But from just the field descriptor typeName doesn't know if we're trying to say "you were trying to set a string on a int (value inside a list)" versus "you were trying to set a string on a field that is a list"

go/protomodule/protomodule_list.go
seena-stripe: I had to look this up https://pkg.go.dev/go.starlark.net/starlark?utm_source=godoc#HasAttrs
> For implementation convenience, a result of (nil, nil) from Attr is interpreted as a "no such field or method" error. Implementations are free to return a more precise error.

kathleen-stripe (approved):

PR #89, opened 2021-08-17
Drop gogo support

seena-stripe:
## Summary
Removes gogo support from Skycfg

In moving to the new go protobuf v2 api, Skycfg will only be able to expose the types in a provided `protoregsitry.Types`, which accepts values https://pkg.go.dev/google.golang.org/protobuf/reflect/protoregistry#Types.RegisterMessage that implement https://pkg.go.dev/google.golang.org/protobuf@v1.27.1/reflect/protoreflect#MessageType. 

However, https://github.com/gogo/protobuf (an alternative go code generator that Skycfg supported) does not support the new API (implementing MessageType and other interfaces). As we move Skycfg to using the new protobuf api, I don't see a way to support gogo and thus this PR drops it.

For any user interested in using Skycfg with gogo, Skycfg was tagged with `v0.1.0` prior to the v2 api switch, and users can continue to use that version with gogo support, but with the old protobuf api.

## Testing
I ran bazel build, test and go build on skycfg and everything compiles (after removing the gogo tests)


README.md
```diff
@@ -175,23 +175,9 @@ Our existing public APIs are expected to be stable even before the v1.0 release.
...
-
-For projects that use `gogo_protobuf`, e.g. envoy, this can be very confusing.
-
-To resolve this, use the package `github.com/stripe/skycfg/gogocompat`
-in your go code. This will allow skycfg code to load
-`proto.package("gogo:google.protobuf") and the internal registry it
-creates will handle the conversion of these types.
+Skycfg has dropped support for
+[gogo/protobuf](https://github.com/gogo/protobuf), an alternative protobuf go
+code generator, when upgrading to using the go-protobuf v2 `pi. For using gogo
```
kathleen-stripe: nit: typo in "`pi" - I'm assuming that's supposed to be "api."

kathleen-stripe (changes requested):
just a small nit

seena-stripe:
Nice, great catch

kathleen-stripe (approved):

PR #90, opened 2021-08-17
Protobuf api V2: add protoMessage, merge and autoboxing wrapperspb 

seena-stripe:
## Summary
This PR adds the remaining type implementation

- `protoMessage` matches https://github.com/stripe/skycfg/blob/trunk/internal/go/skycfg/proto_message.go
- `maybeConvertToWrapper` copies https://github.com/stripe/skycfg/blob/34104506579982852541ef96cef964e179fa767b/internal/go/skycfg/proto_message.go#L368
- `merge` is an implementation of `proto.Merge` 
 

## Implementation differences
The main implementation differences here is the new implementation does not store values on the underlying proto.Message, which is only used as a type to later instantiate when using `toProtoMessage`. All the values are stored only on the `fields` map (which is similar to the `attrCache`) from before but the old implementation attempted to keep both `msg` and `attrCache` in sync, producing a lot of extra code and some weird edge case bugs.

This means that `NewMessage` will not mutate the incoming proto message. This is mostly an internal detail except when passing in global values into skycfg -- the evaluated starlark will no longer be able to mutate incoming global variables. IMO this is a good thing, as it's surprising behavior, and a user can always fetch the updated value with `AsProtoMessage`

Example: 
```go
msg := &pb.MessageV2{}
globals := starlark.StringDict{
	"msg": protomodule.NewMessage(msg),
}
_, err := starlark.Eval(&starlark.Thread{}, "", test.src, globals)
```

In the old implementation, `msg` is directly mutated, so `starlark.Eval` can cause a side effect. This means repeatedly running `eval` can produce different results.

In the new implementation, `msg` is unchanged, and the user can use the following to extract the changed global if they wish
```
wrapped := protomodule.NewMessage(msg)
globals := starlark.StringDict{
	"msg": wrapped,
}
_, err := starlark.Eval(&starlark.Thread{}, "", test.src, globals)
out, err := AsProtoMessage(wrapped)
```

## Tests
This is all tested on https://github.com/stripe/skycfg/tree/seena/protobuf-api-v2 but am not adding the tests as part of this PR as they involve making skycfg use `*protoMessage` in place of the legacy `*skyProtoMessage` -- this is the switch from old to new implementation, and I thought it would be better to separate that into it's own PR, since again this pr is mostly boilerplate

### DIFF
--- a/go/protomodule/merge.go
+++ b/go/protomodule/merge.go
@@ -1,0 +1,97 @@
+// Copyright 2021 The Skycfg Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package protomodule
+
+import (
+	"fmt"
+
+	"go.starlark.net/starlark"
+)
+
+// Implements proto.merge merging src into dst, returning merged value
+func mergeField(dst, src starlark.Value) (starlark.Value, error) {
+	if dst == nil {
+		return src, nil
+	}
+	if src == nil {
+		return dst, nil
+	}
+
+	if dst.Type() != src.Type() {
+		return nil, mergeError(dst, src)
+	}
+
+	switch dst := dst.(type) {
+	case *protoRepeated:
+		if src, ok := src.(*protoRepeated); ok {
+			newList := newProtoRepeated(dst.fieldDesc)
+
+			err := newList.Extend(dst)
+			if err != nil {
+				return nil, err
+			}
+
+			err = newList.Extend(src)
+			if err != nil {
+				return nil, err
+			}
+
+			return newList, nil
+		}
+		return nil, mergeError(dst, src)
+	case *protoMap:
+		if src, ok := src.(*protoMap); ok {
+			newMap := newProtoMap(dst.mapKey, dst.mapValue)
+
+			for _, item := range dst.Items() {
+				err := newMap.SetKey(item[0], item[1])
+				if err != nil {
+					return nil, err
+				}
+			}
+
+			for _, item := range src.Items() {
+				err := newMap.SetKey(item[0], item[1])
+				if err != nil {
+					return nil, err
+				}
+			}
+
+			return newMap, nil
+		}
+		return nil, mergeError(dst, src)
+	case *protoMessage:
+		if src, ok := src.(*protoMessage); ok {
+			newMessage, err := NewMessage(dst.msg)
+			if err != nil {
+				return nil, err
+			}
+
+			newMessage.Merge(dst)
+			newMessage.Merge(src)
+
+			return newMessage, nil
+		}
+		return nil, mergeError(dst, src)
+	default:
+		return src, nil
+	}
+}
+
+func mergeError(dst, src starlark.Value) error {
+	return fmt.Errorf("MergeError: Cannot merge protobufs of different types: Merge(%s, %s)", dst.Type(), src.Type())
+}
--- a/go/protomodule/merge.go
+++ b/go/protomodule/merge.go
@@ -37,56 +37,62 @@ func mergeField(dst, src starlark.Value) (starlark.Value, error) {
 
 	switch dst := dst.(type) {
 	case *protoRepeated:
-		if src, ok := src.(*protoRepeated); ok {
-			newList := newProtoRepeated(dst.fieldDesc)
-
-			err := newList.Extend(dst)
-			if err != nil {
-				return nil, err
-			}
-
-			err = newList.Extend(src)
-			if err != nil {
-				return nil, err
-			}
-
-			return newList, nil
+		src, ok := src.(*protoRepeated)
+		if !ok {
+			return nil, mergeError(dst, src)
 		}
-		return nil, mergeError(dst, src)
+
+		newList := newProtoRepeated(dst.fieldDesc)
+
+		err := newList.Extend(dst)
+		if err != nil {
+			return nil, err
+		}
+
+		err = newList.Extend(src)
+		if err != nil {
+			return nil, err
+		}
+
+		return newList, nil
 	case *protoMap:
-		if src, ok := src.(*protoMap); ok {
-			newMap := newProtoMap(dst.mapKey, dst.mapValue)
-
-			for _, item := range dst.Items() {
-				err := newMap.SetKey(item[0], item[1])
-				if err != nil {
-					return nil, err
-				}
-			}
-
-			for _, item := range src.Items() {
-				err := newMap.SetKey(item[0], item[1])
-				if err != nil {
-					return nil, err
-				}
-			}
-
-			return newMap, nil
+		src, ok := src.(*protoMap)
+		if !ok {
+			return nil, mergeError(dst, src)
 		}
-		return nil, mergeError(dst, src)
-	case *protoMessage:
-		if src, ok := src.(*protoMessage); ok {
-			newMessage, err := NewMessage(dst.msg)
+
+		newMap := newProtoMap(dst.mapKey, dst.mapValue)
+
+		for _, item := range dst.Items() {
+			err := newMap.SetKey(item[0], item[1])
 			if err != nil {
 				return nil, err
 			}
-
-			newMessage.Merge(dst)
-			newMessage.Merge(src)
-
-			return newMessage, nil
 		}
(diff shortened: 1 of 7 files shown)


go/protomodule/merge.go
```diff
@@ -0,0 +1,97 @@
...
+
+	switch dst := dst.(type) {
+	case *protoRepeated:
+		if src, ok := src.(*protoRepeated); ok {
+			newList := newProtoRepeated(dst.fieldDesc)
+
+			err := newList.Extend(dst)
+			if err != nil {
+				return nil, err
+			}
```
kathleen-stripe: nitty suggestion: To make this slightly more concise, you could write

```
if err := newList.Extend(dst); err != nil {
	return nil, err
}
```

This was the preferred syntax at my last employer - not sure if there's any stylistic opinions about this at Stripe.

go/protomodule/merge.go
```diff
@@ -0,0 +1,97 @@
...
+			err = newList.Extend(src)
+			if err != nil {
+				return nil, err
+			}
+
+			return newList, nil
+		}
+		return nil, mergeError(dst, src)
+	case *protoMap:
+		if src, ok := src.(*protoMap); ok {
```
kathleen-stripe: nitty suggestion: This might be slightly more readable if you checked `!ok` instead of `ok` so that the conditional body is shorter and more of it is outdented, e.g.

```
if src, ok := src.(*protoMap); !ok {
  return nil, mergeError(dst, src)
}

newMap := newProtoMap(dst.mapKey, dst.mapValue)
// and the rest of the body (the for loops) goes here
```

go/protomodule/protomodule_list.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 The Skycfg Authors.
+// Copyright 2021 The Skycfg Authors.
```
kathleen-stripe, line 1: 😎

go/protomodule/protomodule_message.go
```diff
@@ -0,0 +1,370 @@
...
+// message. It can be returned back to a proto.Message() via AsProtoMessage().
+//
+// NewMessage copies the input proto.Message and therefore does not modify it
+func NewMessage(msg proto.Message) (*protoMessage, error) {
+	msgDesc := msg.ProtoReflect().Descriptor()
+
+	fields := make(map[string]starlark.Value)
+
+	// Copy any existing set fields
+	msgReflect := msg.ProtoReflect()
```
kathleen-stripe: nit: Since `msg.ProtoReflect()` is also used on line 36, could we move this line to be before line 36 so that line 36 can use this variable?

go/protomodule/protomodule_message.go
```diff
@@ -0,0 +1,370 @@
...
+			}
+			fields[string(fd.Name())] = starlarkValue
+
+		}
+		return true
+	})
+
+	// Clone and reset the input msg to ensure no mutations
+	cloned := proto.Clone(msg)
+	proto.Reset(cloned)
```
kathleen-stripe, line 56: Since we're resetting `cloned`, does that make `cloned` an empty proto message?

go/protomodule/protomodule_message.go
```diff
@@ -0,0 +1,370 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Clear other oneof
+	oneof := fieldDesc.ContainingOneof()
+	if oneof != nil {
+		fields := oneof.Fields()
+		for i := 0; i < fields.Len(); i++ {
+			delete(msg.fields, string(fields.Get(i).Name()))
```
kathleen-stripe, line 257: I don't fully understand the data structure for one of fields - are the fields in `oneof.Fields()` actually expected to be in `msg.fields`? Or are they in some other data structure nested within `msg.fields`? (Feel free to point me at documentation.)

go/protomodule/protomodule_message.go
```diff
@@ -0,0 +1,370 @@
...
+		if _, ok := msg.fields[fieldName]; ok {
+			continue
+		}
+
+		if !fieldDesc.HasDefault() {
+			continue
+		}
+
+		val, err := valueToStarlark(fieldDesc.Default(), fieldDesc)
+		if err != nil {
```
kathleen-stripe, line 288: OOC - would we want to revert the fields for which we've already set defaults? Or is it okay to leave the proto message in a partially updated state where some fields have defaults and others don't (but could)?

Relatedly, is there value in collecting errors (to return at the end of the method) while continuing to set as many default values as possible?

go/protomodule/protomodule_message.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+
+	for fieldName, val := range other.fields {
+		fieldDesc := getFieldDescriptor(msg.msgDesc, fieldName)
+		if fieldDesc == nil {
+			continue
+		}
+
+		merged, err := mergeField(msg.fields[fieldName], val)
+		if err != nil {
```
kathleen-stripe, line 315: same question here about whether it's safe/expected to leave the proto in a partially-merged state

go/protomodule/type_conversions.go
```diff
@@ -204,12 +217,114 @@ func scalarValueToStarlark(val protoreflect.Value, fieldDesc protoreflect.FieldD
...
+	case Int64ValueType:
+		switch val := val.(type) {
+		case starlark.Int:
+			int64Val, ok := val.Int64()
+			if ok {
+				return NewMessage(&wrapperspb.Int64Value{Value: int64Val})
+			}
+			return nil, fmt.Errorf("ValueError: value %v is not exactly representable as type `int64'.", val)
+		}
+	case UInt32ValueType:
```
kathleen-stripe: There's already a check for this case on line 253 (similarly for case `UInt64ValueType` on line 315 below).

EDIT: This is the only blocking comment.

go/protomodule/protomodule_message.go
seena-stripe: Yes, the values are copied onto the `fields` map and this cloned/reset message is only used to later construct a concrete type in `toProtoMessage`.  I wanted to explicitly clear it to make the intent clearer

go/protomodule/merge.go
seena-stripe: I haven't seen any stylistic opinions one way or the other, and I don't have a strong personal preference. That said, if it's okay, I'll leave this as is for these PRs as it seems really inconsistent to change it in some places but not others

go/protomodule/protomodule_message.go
seena-stripe: Are you asking if `SetDefaults` should behave like first clearing the message like `proto.Reset` and then setting the defaults for fields that have them? If so this is what I would expect and that's what the stated behavior in the docs seems to apply https://github.com/stripe/skycfg/blob/trunk/docs/modules.asciidoc#proto.set_defaults but let me look into how the current implementation behaves

With errors, I don't think it really matters as when an error is returned the starlark/skycfg evaluation stops and returns the error. Since Skycfg is not mutating any values outside of the evaluation, all these values will just get thrown away anyway even if in a partial state

go/protomodule/protomodule_message.go
seena-stripe: So my understanding of a `oneof` field is that it has a separate `FieldDescriptor` per field, and `oneof.Fields()` returns all the `FieldDescriptors` connected to the same oneof. In our test proto this would be `f_oneof_a` and `f_oneof_b`: https://github.com/stripe/skycfg/blob/trunk/internal/testdata/test_proto/test_proto_v3.proto#L59-L62

In this implementation, if you set any oneof, it clears the other values from the `fields` map (unsetting them) otherwise when `toProtoMessage` iterates over `fields` and calls `msg.Set` on each of them, it will choose the last oneof value set as only one oneof is set and all others are cleared on set

At least in the dynamicpb code this works the same https://github.com/protocolbuffers/protobuf-go/blob/v1.27.1/types/dynamicpb/dynamic.go#L269

go/protomodule/protomodule_message.go
seena-stripe: Alright I tested the existing implementation and it does not behave this way, only setting values with defaults not clearing those without. It calls out to https://github.com/golang/protobuf/blob/master/proto/defaults.go#L25 and here it's filtering out to only fields with `HasDefault()` which AFAICT is only true for fields with proto2 defaults

This deviates from my expectation/Skycfg's state docs but is also what protobuf is doing as well. That said I would prefer to match the existing behavior as closely as possible to de-risk the new implementation refactor, and then change any behavior separately/explicitly afterwards

kathleen-stripe: Sorry for the confusion!

> Are you asking if SetDefaults should behave like first clearing the message like proto.Reset and then setting the defaults for fields that have them?

No, I was asking "when `SetDefaults` errors, should we clear the errors that were successfully set earlier in `SetDefaults`?" i.e. Should setting defaults be all-or-nothing or "best effort"?

But based on
> Since Skycfg is not mutating any values outside of the evaluation, all these values will just get thrown away anyway even if in a partial state

it sounds like this is all-or-nothing?

>That said I would prefer to match the existing behavior as closely as possible to de-risk the new implementation refactor, and then change any behavior separately/explicitly afterwards

👍 SGTM!

kathleen-stripe (approved):

PR #91, opened 2021-08-17
Support passing in variables into tests

seena-stripe:
## Summary
Add option to pass in variables into executed tests.

Skycfg supports passing in variables during evaluation, but does not have a similar option to do so when evaluating tests. This works fine for tests on helper functions (where you'd have the caller pass relevant values as parameters rather than accessing the global `ctx.vars` directly) but meant you can't test the output of skycfg evaluation.

For example, if I wanted to assert that my proto's `f_string` is only 100 characters long:

```
def main(ctx):
    msg = MessageV3()
    msg.f_string = "hello, " + ctx.vars["name"]
    return [msg]

def test_main(ctx):
    msg = main(ctx)[0]
    ctx.assert(len(main) < 100)
```

Previously this would fail when running `test_main` because `ctx.vars["name"]` is not available. With this pr, it is possible to pass in the same variables when running tests.

## Tests
Updated tests to skycfg_test to consider this case

### DIFF
--- a/internal/go/skycfg/skycfg_test.go
+++ b/internal/go/skycfg/skycfg_test.go
@@ -45,10 +45,16 @@ def test_helper1(t):
 	x = helper1()
 	t.assert(x == 12345)
 
+def test_main(ctx):
+	msg = main(ctx)[0]
+	ctx.assert(len(msg.r_string) == 1)
+	ctx.assert(msg.r_string[0] == "var_value")
+
 def main(ctx):
 	msg = test_proto.MessageV2()
 	msg.f_int64 = helper1()
 	msg.f_string = json.encode(helper2(ctx))
+	msg.r_string.append(ctx.vars["var_key"])
 
 	return [msg]
 `,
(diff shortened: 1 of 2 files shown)


internal/go/skycfg/skycfg_test.go
```diff
@@ -399,10 +401,20 @@ func TestSkycfgTesting(t *testing.T) {
...
 			errors: true,
 		},
+		"test_main": testTestCase{
+			passes: true,
+		},
+	}
+
+	tests := config.Tests()
+	if len(tests) != len(cases) {
+		t.Error("Expected %d tests but found", len(cases), len(tests))
```
kathleen-stripe: I think you're missing the placeholder for the second interpolated value

skycfg.go
```diff
@@ -436,9 +436,39 @@ func (t *Test) Name() string {
...
+
+type testOptions struct {
+	vars *starlark.Dict
+}
+
+type fnTestOption func(*testOptions)
+
+func (fn fnTestOption) applyTest(opts *testOptions) { fn(opts) }
+
+// WithVars adds key:value pairs to the ctx.vars dict passed to tests
```
kathleen-stripe: nit: should be `WithTestVars`?

kathleen-stripe (dismissed):

kathleen-stripe (approved):

PR #93, opened 2021-08-18
Protobuf API V2: finish implementation and switchover skycfg to using new impl

seena-stripe:
**This PR is currently stacked on top of #90, please review by commit**

## Summary
This PR completes the implementation on the new go protobuf v2 api, and makes the cutover to using the new implementation (points skycfg.go at go/protomodule, deletes internal/go/skycfg).

Specifically this pr
- Changes `protomodule.go`'s global functions (e.g. `proto.clear`) to operate on the new `*protoMessage` object rather than the old message type.
- Adds `package` to `protomodule` and finishes implementing `message_type` so that `pkg = proto.package("skycfg.test_proto")` will return an object representing the package, and then `pkg.MessageV3` returns an object (`protoMessageType`)  (that can construct a new `*protoMessage` with `MessageV3` 
- With global functions, package and message_type working, this PR adds tests on the message object and global functions
- Deletes `internal/go/skycfg`, but moves `internal/go/skycfg/skycfg_test.go` to the top level directory
- Updates `skcyfg.go` to use the new `go/protomodule`

## Test
This PR adds the original `internal/go/skycfg/proto_test.go` and `internal/go/skycfg/skycfg_test.go` to testing the new implementation

I have also tested this internally on our existing usage of skycfg, snapshotting before and after the implementation switch and confirming there are no differences in output (at least in our usage)


go/protomodule/protomodule_message_test.go
```diff
@@ -0,0 +1,959 @@
...
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			_, err := starlark.Eval(&starlark.Thread{}, "", `pb.`+test.src, globals)
+			if test.wantErr != "" {
+				if err == nil {
+					t.Errorf("eval(%q): expected error", test.src)
+				} else if test.wantErr != err.Error() {
+					t.Errorf("eval(%q): expected error \n%q\n%q", test.src, test.wantErr, err.Error())
```
kathleen-stripe: nit: might be a bit clearer if the error message labeled the error that was received, e.g.

```
"eval(%q): expected error \n%q\ngot:\n%q"
```

go/protomodule/protomodule_message_test.go
```diff
@@ -0,0 +1,959 @@
...
+		RString: []string{"a", "b", "c"},
+	}
+	skyMsg, _ := NewMessage(msg)
+
+	// create a separate msg to ensure the underlying reference in skyMsgOther is different
+	msgOther := &pb.MessageV2{
+		RString: []string{"a", "b"},
+	}
+	skyMsgOther, _ := NewMessage(msgOther)
+	ok, err := starlark.Compare(syntax.EQL, skyMsg, skyMsgOther)
```
kathleen-stripe, line 555: should this use `syntax.NEQL`?

go/protomodule/protomodule_message_test.go
```diff
@@ -0,0 +1,959 @@
...
+
+func TestProtoMergeDiffTypes(t *testing.T) {
+	errorMsg := "proto.merge: types are not the same: got skycfg.test_proto.MessageV3 and skycfg.test_proto.MessageV2"
+	globals := starlark.StringDict{
+		"proto": NewModule(newRegistry()),
+	}
+	src, err := starlark.Eval(&starlark.Thread{}, "",
+		`proto.merge(proto.package("skycfg.test_proto").MessageV2(), proto.package("skycfg.test_proto").MessageV3())`, globals)
+	if err == nil {
+		t.Errorf("expected error, got %q", src)
```
kathleen-stripe: nit: might be useful to include the expected error here?

go/protomodule/protomodule_message_test.go
kathleen-stripe: This is the only blocking comment.

seena-stripe: I think this test (copied) was testing `EQL` returns false, but I added another clause here to check `NEQ` returns true as well

kathleen-stripe: OH. I misread it the first time - sorry!

kathleen-stripe (approved):

PR #94, opened 2021-08-20
Protobuf API V2: compatibility fixes

seena-stripe:
## Summary
Two fixes to maintain compatibility with the old implementation

- **Add json module members for compat** (852f982):
  `json.marshal` became `json.encode`. Added these aliases back like we do for `yamlModule`
- **Allow constructing message maps with None** (416b96c): 
  Skycfg previously allowed setting proto2 message values with None, effectively unsetting them. For maps, this set them to be nil, so
```python
# skycfg
msg.map_submsg = { 
  "a": None
}
```
became
```go
&pb.MessageV2{
  map_submsg: map[string]*pb.MessageV2{
    "a": nil,
  }
}
```
With protoreflect, `map.Set` only accepts a `protoreflect.Value` https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect#Map.Set (a struct, cannot be nil) so my two options are 1) create an empty `&pb.MessageV2{}` or treat `None` as not setting the key. I went with the latter, to make it behave like `protoMessage` but neither option seems _great_

- **Allow set to None in proto2 mode for scalar values** (873051b): this adds back the behavior from https://github.com/stripe/skycfg/blob/e785c6f705dd143a6756f2a0f56005ae824962d4/internal/go/skycfg/proto_test.go#L1069 and  https://github.com/stripe/skycfg/blob/e785c6f705dd143a6756f2a0f56005ae824962d4/internal/go/skycfg/proto_test.go#L870

## Tests
I added a test for compatibility

### DIFF
--- a/go/protomodule/protomodule_map.go
+++ b/go/protomodule/protomodule_map.go
@@ -60,11 +60,11 @@ func newProtoMapFromDict(mapKey protoreflect.FieldDescriptor, mapValue protorefl
 	out := &protoMap{
 		mapKey:   mapKey,
 		mapValue: mapValue,
-		dict:     d,
+		dict:     starlark.NewDict(d.Len()),
 	}
 
 	for _, item := range d.Items() {
-		err := out.typeCheck(item[0], item[1])
+		err := out.SetKey(item[0], item[1])
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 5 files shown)


go/protomodule/protomodule_map.go
```diff
@@ -145,19 +145,25 @@ func (m *protoMap) wrapUpdate() starlark.Value {
...
+	err := scalarTypeCheck(m.mapKey, k)
 	if err != nil {
 		return err
 	}
 
-	return m.dict.SetKey(k, v)
-}
+	// Pre 1.0 compatibility allowed maps to be constructed with None in proto2
+	// with the value treated as nil. protoreflect does not allow setting
+	// with a value of nil, so instead treat it as an unset
```
kathleen-stripe, line 156: thanks for including this comment!

go/protomodule/protomodule_message_test.go
```diff
@@ -966,3 +967,82 @@ func TestProtoMergeDiffTypes(t *testing.T) {
...
+	}
+	got := mustProtoMessage(t, val).(*pb.MessageV2)
+
+	checkProtoEqual(t, &pb.MessageV2{
+		MapSubmsg: map[string]*pb.MessageV2{
+			"a": &pb.MessageV2{},
+		},
+	}, got)
+
+	// Confirm this only works for all in proto2, only message values in proto3
```
kathleen-stripe, line 1003: tbh, I'm not sure if I'm able to parse this comment (line 1003) correctly, but I'm assuming it says something like "confirm that setting a map value to None works for proto2 but not for proto3"?

kathleen-stripe (approved):

go/protomodule/protomodule_message_test.go
seena-stripe: I believe the intent in the previous implementation is that scalar and message values can be set to none in proto2, but in proto3 only setting messages to none is allowed (equivalent to unsetting).

With the behavior now being to unset the fields, it seems sort of arbitrary to allow it in proto2 versus proto3 but just trying to capture the original behavior. IMO ideally we wouldn't support setting none in maps to unset

PR #95, opened 2021-08-25
Add tests to ensure no copy on assignment, fix map

seena-stripe:
## Summary
This PR adds tests to ensure `protoMap`, `protoList` and `protoMessage` are not copying values on assignment and corrects `protoMap`. Yesterday I learned Skycfg previously did this _sometimes_, e.g. in the old implementation this would happen:

```python
# v0.1.0 Behavior
msg1.r_string = ["a","b"]
a = msg1.r_string
msg2.r_string = msg1.r_string
a.append("c")

# a = [a, b, c]
# msg1.r_string = [a, b, c]
# msg2.r_string = [a, b]
```
(https://github.com/stripe/skycfg/blob/677b39f98fc6399c738a295fc3d77690ffda54c7/internal/go/skycfg/proto_message.go#L518)

The re-implementation fixed this for lists but not for maps incidentally so I added tests to make this behavior explicit. AFAICT from testing starlark, starlark does not copy on assignment for lists and dictionaries (like python) and seems to be what is described in https://github.com/bazelbuild/starlark/blob/master/spec.md#value-concepts

## Tests
Added tests when assigning skycfg and starlark values into a skycfg object

### DIFF
--- a/go/protomodule/protomodule_map.go
+++ b/go/protomodule/protomodule_map.go
@@ -74,6 +74,16 @@ func newProtoMapFromDict(mapKey protoreflect.FieldDescriptor, mapValue protorefl
 		}
 	}
 
+	// Remove any None values from map, see SetKey for compability behavior
+	for _, item := range d.Items() {
+		if item[1] == starlark.None {
+			_, _, err := d.Delete(item[0])
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
 	return out, nil
 }
 
--- a/go/protomodule/protomodule_map.go
+++ b/go/protomodule/protomodule_map.go
@@ -60,11 +60,15 @@ func newProtoMapFromDict(mapKey protoreflect.FieldDescriptor, mapValue protorefl
 	out := &protoMap{
 		mapKey:   mapKey,
 		mapValue: mapValue,
-		dict:     starlark.NewDict(d.Len()),
+		dict:     d,
 	}
 
+	// SetKey is used to typecheck fields appropriately but done on a temporary object
+	// so that the underlying out.dict still has a reference to the given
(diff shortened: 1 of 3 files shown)


go/protomodule/protomodule_map.go
```diff
@@ -60,16 +60,30 @@ func newProtoMapFromDict(mapKey protoreflect.FieldDescriptor, mapValue protorefl
...
+	tmpMap := newProtoMap(mapKey, mapValue)
 	for _, item := range d.Items() {
-		err := out.SetKey(item[0], item[1])
+		err := tmpMap.SetKey(item[0], item[1])
 		if err != nil {
 			return nil, err
 		}
 	}
 
+	// Remove any None values from map, see SetKey for compability behavior
```
kathleen-stripe: nit: `compability` --> `compatibility`?

kathleen-stripe (dismissed):
LGTM - one non-blocking comment. Thanks for fixing this behavior and adding test coverage!

go/protomodule/protomodule_map.go
seena-stripe: Oof good catch

kathleen-stripe (approved):

com6056:
@seena-stripe you mentioned that the old implementation https://github.com/stripe/skycfg/blob/677b39f98fc6399c738a295fc3d77690ffda54c7/internal/go/skycfg/proto_message.go#L518 still has this issue, unfortunately we can't upgrade to the new implementation quite yet due to k8s still using gogoproto.

Do you happen to know how to fix this in the old implementation? I've tried a few things and couldn't quite get it working, was curious if you already knew how to solve it though.

PR #98, opened 2021-08-31
Simplify tests with runSkycfgTests test runner

seena-stripe:
## Summary
Adds a test runner `runSkycfgTests` to remove redundant testing logic. Many of the tests had a 1 line skycfg expression and a 1 line desired output, but duplicated all the test boilerplate, so this PR adds a test runner to simplify setting up a new test

This is purely cleanup and shouldn't affect evaluation.

Example
```diff
 func TestUnsetProto2Fields(t *testing.T) {
        // Proto v2 distinguishes between unset and set-to-empty.
-       msg, err := eval(`proto.package("skycfg.test_proto").MessageV2(
-                f_string = None,
-        )`, nil)
-       if err != nil {
-               t.Fatal(err)
-       }
-       got := mustProtoMessage(t, msg)
-       want := &pb.MessageV2{
-               FString: nil,
-       }
-       checkProtoEqual(t, want, got)
+       runSkycfgTests(t, []skycfgTest{
+               {
+                       src: `proto.package("skycfg.test_proto").MessageV2(
+                               f_string = None,
+                       )`,
+                       want: &pb.MessageV2{
+                               FString: nil,
+                       },
+               },
+       })
 }
```

This doesn't save on LoC very much but reduces the boilerplate so the test code is a lot more signalful and test output is consistent

### DIFF
--- a/go/protomodule/protomodule_message_test.go
+++ b/go/protomodule/protomodule_message_test.go
@@ -380,7 +380,7 @@ func TestAttrValidation(t *testing.T) {
 		"pb": NewProtoPackage(newRegistry(), "skycfg.test_proto"),
 	}
 
-	runSkycfgTests(t, []skycfgTest{
+	tests := []skycfgTest{
 		// Scalar type mismatch
 		{
 			name:    "int32",
@@ -535,7 +535,8 @@ func TestAttrValidation(t *testing.T) {
 			src:     fmt.Sprintf(`pb.MessageV3(f_uint64 = %d - 1)`, 0),
 			wantErr: fmt.Errorf(`ValueError: value -1 overflows type "uint64".`),
 		},
-	}, withGlobals(globals))
+	}
+	runSkycfgTests(t, tests, withGlobals(globals))
 }
 
 func TestProtoMessageString(t *testing.T) {
--- a/go/protomodule/protomodule_message_test.go
+++ b/go/protomodule/protomodule_message_test.go
@@ -376,237 +376,201 @@ func TestMessageV3(t *testing.T) {
 }
 
 func TestAttrValidation(t *testing.T) {
-	tests := []struct {
-		name    string
-		src     string
-		wantErr string
-	}{
+	globals := starlark.StringDict{
+		"pb": NewProtoPackage(newRegistry(), "skycfg.test_proto"),
+	}
 
+	runSkycfgTests(t, []skycfgTest{
 		// Scalar type mismatch
 		{
 			name:    "int32",
-			src:     `MessageV3(f_int32 = '')`,
-			wantErr: `TypeError: value "" (type "string") can't be assigned to type "int32".`,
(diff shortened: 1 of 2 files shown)


go/protomodule/protomodule_message_test.go
```diff
@@ -288,237 +288,201 @@ func TestMessageV3(t *testing.T) {
...
-	tests := []struct {
-		name    string
-		src     string
-		wantErr string
-	}{
+	globals := starlark.StringDict{
+		"pb": NewProtoPackage(newRegistry(), "skycfg.test_proto"),
+	}
 
+	runSkycfgTests(t, []skycfgTest{
```
kathleen-stripe: nit: This function invocation might be easier to read if the `[]skycfgTest` slice was extracted into a variable and the variable was passed to `runSkycfgTests` instead.

go/protomodule/protomodule_test.go
```diff
@@ -799,6 +603,87 @@ func TestProtoToAnyV3(t *testing.T) {
...
+			case proto.Message:
+				got := mustProtoMessage(t, val)
+				checkProtoEqual(t, want, got)
+			case string:
+				got := val.String()
+				if test.removeRandomSpace {
+					got = removeRandomSpace(got)
+				}
+				if want != got {
+					t.Fatalf("Expected\nwanted: %s\ngot   : %s", want, got)
```
kathleen-stripe: nit: not sure if this error makes sense - could we remove the "Expected" or reword? (same for lines below)

EDIT: Or is "Expected" intended to indicate what expectation failed? e.g. expected value vs expected error?

go/protomodule/protomodule_test.go
```diff
@@ -840,11 +725,14 @@ func mustProtoMessage(t *testing.T, v starlark.Value) proto.Message {
...
 
-func checkError(got, want error) bool {
-	if got == nil {
-		return false
+func checkError(t *testing.T, got, want error) {
+	t.Helper()
+
+	if want == nil && got != nil {
+		t.Fatalf("Expected no error, got: %v\n", got)
+	} else if want != nil && got.Error() != want.Error() {
```
kathleen-stripe, line 735: It seems like we also need to handle `want != nil && got == nil`, right?

kathleen-stripe (changes requested):
Only the `want != nil && got == nil` change is blocking. Thanks for doing this refactor!

go/protomodule/protomodule_test.go
seena-stripe: Yeah it was intended to be "Value received differed from expected" or something similar, but I'm not sure how to word it best and it's not providing much value over `wanted: foo, got: bar` so I'm just removing it

go/protomodule/protomodule_test.go
seena-stripe: Oh yeah, great catch

kathleen-stripe (approved):

PR #99, opened 2021-09-03
valueToStarlark support for enumKind

seena-stripe:
## Summary

`valueToStarlark` was missing support for enumKind so this PR:

- Return an error if a `NewMessage` fails to assign a value, updating the type signature in `skycfg.go`
- Adds a test for `msg == AsProtoMessage(NewMessage(msg))` which catches the issue
- Add missing support in `valueToStarlark`

## Tests
Added tests which should cover issues like this in the future

### DIFF
--- a/skycfg.go
+++ b/skycfg.go
@@ -86,9 +86,8 @@ func (r *localFileReader) ReadFile(ctx context.Context, path string) ([]byte, er
 
 // NewProtoMessage returns a Starlark value representing the given Protobuf
 // message. It can be returned back to a proto.Message() via AsProtoMessage().
-func NewProtoMessage(msg proto.Message) starlark.Value {
-	v, _ := protomodule.NewMessage(msg)
-	return v
+func NewProtoMessage(msg proto.Message) (starlark.Value, error) {
+	return protomodule.NewMessage(msg)
 }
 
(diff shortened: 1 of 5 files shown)


skycfg.go
```diff
@@ -86,9 +86,8 @@ func (r *localFileReader) ReadFile(ctx context.Context, path string) ([]byte, er
 
 // NewProtoMessage returns a Starlark value representing the given Protobuf
 // message. It can be returned back to a proto.Message() via AsProtoMessage().
-func NewProtoMessage(msg proto.Message) starlark.Value {
-	v, _ := protomodule.NewMessage(msg)
-	return v
+func NewProtoMessage(msg proto.Message) (starlark.Value, error) {
```
kathleen-stripe, line 89: Since this is a backwards-incompatible change, are you going to release these changes as a new (major?) version for skycfg?

kathleen-stripe (approved):

skycfg.go
seena-stripe: Yes, as I understand the current state is Skycfg has a single release tagged 0.1.0 and there have been a number of other API changes. Also as we learned in https://github.com/stripe/skycfg/pull/95 API changes are easier to deal with than implementation changing untested behavior

PR #100, opened 2021-09-03
Fix autoboxing of wrapperspb types

seena-stripe:
## Summary
Autoboxing of wrapperspb types was only partially working, allowing `msg.f_StringValue = "something"` to typecheck (typechecking was done by `valueFromStarlark` which does the autoboxing) but the assignment was actually assigning `"something"` a string into the internal `fields` map. Then when access happens, `msg.f_StringValue.value` fails because `msg.f_StringValue == "something"` and does not have `.value`

This PR removes `maybeConvertToWrapper` from `valueFromStarlark` so typechecking fails if the conversion has not been explicitly done before assignment. The conversion is moved earlier into `protoMessage.SetField` where the `list/dict` conversions are done.

As an extra fun, there was existing support for converting `["s1", "s2"]` into `[StringValue{value: s1}, StringValue{value: s2}]` which breaks the list assignment semantics. In this case the assignment mutates like map do for None, though it could alternatively copy if we want.

Longer term, I'd really like to get rid of this autoboxing behavior, its convenient but adds a lot of complexity

## Tests
I added tests to `protomodule_message_test`

### DIFF
--- a/internal/testdata/test_proto/test_proto_v2.proto
+++ b/internal/testdata/test_proto/test_proto_v2.proto
@@ -19,6 +19,8 @@ syntax = "proto2";
 option go_package = "github.com/stripe/skycfg/internal/testdata/test_proto";
 package skycfg.test_proto;
 
+import "google/protobuf/wrappers.proto";
+
 message MessageV2 {
   optional int32  f_int32   = 1;
   optional int64  f_int64   = 2;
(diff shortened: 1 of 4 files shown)


internal/testdata/test_proto/test_proto_v2.proto
```diff
@@ -60,7 +62,18 @@ message MessageV2 {
...
+  optional google.protobuf.BoolValue f_BoolValue = 20;
+  optional google.protobuf.StringValue f_StringValue = 21;
+  optional google.protobuf.DoubleValue f_DoubleValue = 22;
+  optional google.protobuf.Int32Value f_Int32Value = 23;
+  optional google.protobuf.Int64Value f_Int64Value = 24;
+  optional google.protobuf.BytesValue f_BytesValue = 25;
+  optional google.protobuf.UInt32Value f_Uint32Value = 26;
+  optional google.protobuf.UInt64Value f_Uint64Value = 27;
+
+  repeated google.protobuf.StringValue r_StringValue = 28;
```
kathleen-stripe, line 74: Should this be `optional`?

kathleen-stripe (changes requested):
LGTM except for possibly one comment - feel free to send back to me for re-review if no changes are needed

internal/testdata/test_proto/test_proto_v2.proto
seena-stripe: I'm not super familiar with proto2 but I think a field can only be one of `optional`, `required` or `repeated` https://developers.google.com/protocol-buffers/docs/proto

kathleen-stripe: Under "[Updating a Message Type](https://developers.google.com/protocol-buffers/docs/proto#updating)," it says
> For string, bytes, and message fields, optional is compatible with repeated

So I think is should be okay to make this field `optional` as well.

kathleen-stripe: nvm - I think I misunderstood the docs. 

EDIT: Further up in that section of the docs, it says:

> Any new fields that you add should be optional or repeated 

My apologies!

kathleen-stripe (approved):

PR #101, opened 2022-01-25
Fixing error message for custom entrypoint

henryc-stripe:
# Summary

Skycfg's entrypoint is usually a `main()` function, but users are allowed to specify a custom function name if they wish. However, error messages always have the entrypoint hardcoded to "main", which is confusing for users specifying a custom entrypoint.

For example, if a user has a custom entrypoint function named `my_entrypoint` and uses the option `skycfg.WithEntryPoint("my_entryponit")` (which has a typo), they will be presented with the error message `no 'main' function found in "file.sky"`, while a more appropriate error message would be `no 'my_entryponit' function found in "file.sky"`. This PR changes the error message so it always uses the entrypoint provided by the user.

### DIFF
--- a/skycfg.go
+++ b/skycfg.go
@@ -382,8 +382,8 @@ func WithEntryPoint(name string) ExecOption {
 	})
 }
 
-// Main executes main() from the top-level Skycfg config module, which is
-// expected to return either None or a list of Protobuf messages.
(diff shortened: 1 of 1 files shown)


seena-stripe (approved):
LGTM!

PR #104, opened 2023-01-30
Add flatten on eval

garvinp-stripe:
Add ability to flatten nested list on evaluations with WithFlattenLists option.
Fix build due to deprecated zlib 1.2.11 
https://github.com/bazelbuild/rules_proto/pull/117

### DIFF
--- a/skycfg.go
+++ b/skycfg.go
@@ -358,8 +358,9 @@ type ExecOption interface {
 }
 
 type execOptions struct {
-	vars     *starlark.Dict
-	funcName string
+	vars         *starlark.Dict
+	funcName     string
+	flattenLists bool
 }
 
 type fnExecOption func(*execOptions)
@@ -379,6 +380,13 @@ func WithVars(vars starlark.StringDict) ExecOption {
 func WithEntryPoint(name string) ExecOption {
 	return fnExecOption(func(opts *execOptions) {
 		opts.funcName = name
+	})
+}
+
+// WithFlattenLists flatten lists one layer deep ([1, [2,3]] -> [1, 2, 3])
+func WithFlattenLists() ExecOption {
+	return fnExecOption(func(opts *execOptions) {
+		opts.flattenLists = true
 	})
 }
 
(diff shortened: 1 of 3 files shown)


skycfg.go
```diff
@@ -426,11 +426,26 @@ func (c *Config) Main(ctx context.Context, opts ...ExecOption) ([]proto.Message,
 	var msgs []proto.Message
 	for ii := 0; ii < mainList.Len(); ii++ {
 		maybeMsg := mainList.Index(ii)
-		msg, ok := AsProtoMessage(maybeMsg)
-		if !ok {
-			return nil, fmt.Errorf("%q returned something that's not a protobuf (a %s)", parsedOpts.funcName, maybeMsg.Type())
+		// Only flatten but not flatten deep. This will flatten out, in order, lists within main list and append the
```
jwog-stripe, line 437: To avoid breaking any other existing uses of skycfg... would you be against making an ExecOption to turn on list flattening? I imagine just add a `flattenLists bool` to the execOptions struct, and then a:
```
// WithFlattenLists flattens lists one layer deep ([1, [2,3]] -> [1, 2, 3])
func WithFlattenLists() ExecOption {
	return fnExecOption(func(opts *execOptions) {
		opts.flattenLists = true
	})
}
```

Then you can case the logic below on `parsedOpts.flattenLists`

skycfg_test.go
```diff
@@ -326,6 +346,24 @@ func TestSkycfgEndToEnd(t *testing.T) {
 				},
 			},
 		},
+		endToEndTestCase{
+			caseName:   "flatten nested list",
+			fileToLoad: "test13.sky",
+			expProtos: []proto.Message{
+				&pb.MessageV2{
```
jwog-stripe, line 354: then add a `flattenLists` field to endToEndTestCase which passes `WithFlattenLists()` as an argument to `config.Main()` and set it to true here

skycfg.go
garvinp-stripe: updated

jwog-stripe (dismissed):
lgtm

skycfg_test.go
```diff
@@ -349,7 +389,12 @@ func TestSkycfgEndToEnd(t *testing.T) {
 	}
 
 	fnExecSkycfg := ExecSkycfg(func(config *skycfg.Config, testCase endToEndTestCase) ([]proto.Message, error) {
-		return config.Main(context.Background(), skycfg.WithVars(testCase.vars))
+		if testCase.flattenLists {
+			return config.Main(context.Background(), skycfg.WithVars(testCase.vars), skycfg.WithFlattenLists())
+		} else {
+			return config.Main(context.Background(), skycfg.WithVars(testCase.vars))
+		}
+
```
dl-stripe: Can we make the `flattenLists` property in the test cases an execOptions slice. We can then provide `skycfg.WithFlattenLists()` as a value in the test case and pass in the `testCase.execOptions...` in the call and forego the if/else statement.

```go
type endToEndTestCase struct {
        ...
	execOptions []skycfg.ExecOption
}


testCases := []endToEndTestCase{
	...
	endToEndTestCase{
			caseName:   "flatten nested list",
			...
			execOptions: []skycfg.ExecOption{skycfg.WithFlattenLists()},
	},
}


return config.Main(context.Background(), skycfg.WithVars(testCase.vars), testCase.execOptions...)
```

dl-stripe (approved):

PR #108, opened 2023-06-10
Make prototext output deterministic

pcj:
Both the protojson and prototext renderings intentionally randomize the whitespace between attributes.  To make it possible to use printed representations of these proto messages in golden tests, we want to make it deterministic for normal daily use and accept that future library upgrades may need to also update the golden files.


pcj:
Closing, this did not fix the issue.

PR #110, opened 2023-08-28
Allow users to configure where print() output goes

timothyg-stripe:
This is necessary for true parallel execution of skycfg, without common access to global state.

Also introduce a `skycfg.CommonOption` type and infrastructure, for options that apply to all three entrypoints for Starlark evaluation: loading, execution, testing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,11 +3,7 @@ module github.com/stripe/skycfg
 go 1.16
 
 require (
-	github.com/golang/protobuf v1.4.1
-	github.com/kylelemons/godebug v0.0.0-20170820004349-d65d576e9348
 	go.starlark.net v0.0.0-20201204201740-42d4f566359b
 	google.golang.org/protobuf v1.25.0
 	gopkg.in/yaml.v2 v2.2.1
 )
-
-replace github.com/kylelemons/godebug => github.com/jmillikin-stripe/godebug v0.0.0-20180620173319-8279e1966bc1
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -3,11 +3,7 @@ module github.com/stripe/skycfg
 go 1.16
 
 require (
-	github.com/golang/protobuf v1.4.1
-	github.com/kylelemons/godebug v0.0.0-20170820004349-d65d576e9348
```
timothyg-stripe, line 7: These are the result of running `go mod tidy`: it appears that we don't use these two deps anywhere.

skycfg.go
```diff
@@ -185,13 +226,13 @@ func WithProtoRegistry(r unstableProtoRegistryV2) LoadOption {
...
-//  * struct - experimental Starlark struct support.
-//  * yaml   - same as "json" package but for YAML.
-//  * url    - utility package for encoding URL query string.
+//   - fail   - interrupts execution and prints a stacktrace.
+//   - hash   - supports md5, sha1 and sha245 functions.
+//   - json   - marshals plain values (dicts, lists, etc) to JSON.
+//   - proto  - package for constructing Protobuf messages.
+//   - struct - experimental Starlark struct support.
+//   - yaml   - same as "json" package but for YAML.
+//   - url    - utility package for encoding URL query string.
```
timothyg-stripe, line 235: This is the result of running `go fmt`: https://go.dev/doc/go1.19#go-doc

aliaqel-stripe (approved):

dl-stripe (approved):

PR #120, opened 2024-06-25
Support name and number accessors on enum values

LINKIWI:
Currently, Protobuf enum values have a `String()` method defined, which provides a human-readable representation of the value with `str(...)` in Starlark, including the value's name and number (index). However, there is no interface to retrieve *only* the name and number of an enum value for use in Skycfg programs.

This PR proposes augmenting the enum value type with two attributes, `name` and `value`, which expose the enum value descriptor's name and number, respectively.

This enables a definition like:

```protobuf
enum Foo {
    HELLO = 0;
    WORLD = 1;
}
```

to be used in Skycfg like:

```starlark
pb.Foo.HELLO.name  # "HELLO" (string)
pb.Foo.HELLO.value  # 0 (int)
```

I added unit tests to cover the new logic (as well as some additional tests to cover existing logic for the `str(...)` representation of values).

```bash
$ bazelisk test --cache_test_results=no //...
INFO: Invocation ID: 0c384b38-64b0-461b-985b-ed281b040175
INFO: Analyzed 14 targets (0 packages loaded, 1 target configured).
INFO: Found 8 targets and 6 test targets...
INFO: Elapsed time: 0.102s, Critical Path: 0.03s
INFO: 7 processes: 1 internal, 6 linux-sandbox.
INFO: Build completed successfully, 7 total actions
//:skycfg_test                                                           PASSED in 0.0s
//go/assertmodule:assertmodule_test                                      PASSED in 0.0s
//go/hashmodule:hashmodule_test                                          PASSED in 0.0s
//go/protomodule:protomodule_test                                        PASSED in 0.0s
//go/urlmodule:urlmodule_test                                            PASSED in 0.0s
//go/yamlmodule:yamlmodule_test                                          PASSED in 0.0s

Executed 6 out of 6 tests: 6 tests pass.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
```

### DIFF
--- a/go/protomodule/protomodule_enum.go
+++ b/go/protomodule/protomodule_enum.go
@@ -109,7 +109,7 @@ func (v *protoEnumValue) Attr(attrName string) (starlark.Value, error) {
 	switch attrName {
 	case "name":
 		return starlark.String(v.value.Name()), nil
-	case "number":
+	case "value":
 		return starlark.MakeInt64(int64(v.value.Number())), nil
 	default:
 		return nil, fmt.Errorf("unknown attribute %s on %s", attrName, v.String())
@@ -117,7 +117,7 @@ func (v *protoEnumValue) Attr(attrName string) (starlark.Value, error) {
 }
 
 func (v *protoEnumValue) AttrNames() []string {
-	return []string{"name", "number"}
+	return []string{"name", "value"}
 }
 
 func (v *protoEnumValue) enumNumber() protoreflect.EnumNumber {
--- a/go/protomodule/protomodule_enum.go
+++ b/go/protomodule/protomodule_enum.go
@@ -92,6 +92,7 @@ type protoEnumValue struct {
 }
 
 var _ starlark.Comparable = (*protoEnumValue)(nil)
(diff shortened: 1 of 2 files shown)


go/protomodule/protomodule_enum.go
```diff
@@ -104,6 +105,21 @@ func (v *protoEnumValue) Hash() (uint32, error) {
 	return starlark.MakeInt64(int64(v.value.Number())).Hash()
 }
 
+func (v *protoEnumValue) Attr(attrName string) (starlark.Value, error) {
+	switch attrName {
+	case "name":
+		return starlark.String(v.value.Name()), nil
+	case "number":
```
sushain-stripe: Should we call this `value`? That seems to be what the [proto docs](https://protobuf.dev/programming-guides/enum/) reference it as.

LINKIWI: Sounds good. Updated

sushain-stripe (approved):

PR #122, opened 2024-09-03
Add murmur3 to hash module

aadi-stripe:
As per the description, we wanted to use murmur3, hence we'd like to add it to the hash module. We are experimenting with it internally at Stripe for some CI tooling, given its known better performance for fair distribution of a small set of inputs.

### DIFF
--- a/go/hashmodule/hashmodule.go
+++ b/go/hashmodule/hashmodule.go
@@ -24,17 +24,20 @@ import (
 	"fmt"
 	"hash"
 
+	"github.com/spaolacci/murmur3"
+
 	"go.starlark.net/starlark"
 	"go.starlark.net/starlarkstruct"
 )
 
 // NewModule returns a Starlark module of common hash functions.
 //
-//  hash = module(
-//    md5,
-//    sha1,
-//    sha256,
-//  )
+// hash = module(
+//	md5,
+//	sha1,
+//	sha256,
+//	murmur3,
+// )
 //
 // See `docs/modules.asciidoc` for details on the API of each function.
 func NewModule() *starlarkstruct.Module {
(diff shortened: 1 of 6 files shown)


go/hashmodule/hashmodule.go
```diff
@@ -24,26 +24,32 @@ import (
...
-//    md5,
-//    sha1,
-//    sha256,
-//  )
+// hash = module(
+//
+//	  md5,
+//		sha1,
+//		sha256,
+//		murmur3,
```
sushain97: Indentation looks off here?

go/hashmodule/hashmodule.go
```diff
@@ -24,26 +24,32 @@ import (
...
 	return &starlarkstruct.Module{
 		Name: "hash",
 		Members: starlark.StringDict{
-			"md5":    starlark.NewBuiltin("hash.md5", fnHash(md5.New)),
-			"sha1":   starlark.NewBuiltin("hash.sha1", fnHash(sha1.New)),
-			"sha256": starlark.NewBuiltin("hash.sha256", fnHash(sha256.New)),
+			"md5":        starlark.NewBuiltin("hash.md5", fnHash(md5.New)),
+			"sha1":       starlark.NewBuiltin("hash.sha1", fnHash(sha1.New)),
+			"sha256":     starlark.NewBuiltin("hash.sha256", fnHash(sha256.New)),
+			"murmur3_64": starlark.NewBuiltin("hash.murmur3_64", fnHash64(murmur3.New64)),
```
sushain97: `murmur3` above but `murmur3_64` here. Former seems better to me.

sushain97 (changes requested):

go/hashmodule/hashmodule.go
aadi-stripe: Ah yeah that's my formatter. Can fix

aadi-stripe: fixed

go/hashmodule/hashmodule.go
aadi-stripe: fixed

sushain97 (approved):

sushain-stripe (approved):

PR #123, opened 2024-09-23
Make protobuf map --> starlark.Dict ordering deterministic

ryanpbrewster:
If I have a protobuf map in the host environment (say as a context variable), iterating through it is non-deterministic. That results in some pretty gnarly workarounds in the starlark code itself.

This PR performs some best-effort deterministic at the starlark conversion stage. It ensures that protobuf map items are inserts into the starlark dictionary in naturally sorted order (using starlark's own internal partial ordering).

Sorting is really awkward in starlark itself, so this type of convenience is extremely helpful

### DIFF
--- a/go/protomodule/type_conversions.go
+++ b/go/protomodule/type_conversions.go
@@ -21,8 +21,10 @@ package protomodule
 import (
 	"fmt"
 	"math"
+	"sort"
 
 	"go.starlark.net/starlark"
+	"go.starlark.net/syntax"
 	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/types/known/anypb"
 	"google.golang.org/protobuf/types/known/wrapperspb"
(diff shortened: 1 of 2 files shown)


go/protomodule/type_conversions.go
```diff
@@ -194,6 +167,73 @@ func valueToStarlark(val protoreflect.Value, fieldDesc protoreflect.FieldDescrip
 	return scalarValueToStarlark(val, fieldDesc)
 }
 
+// Wrap a protobuf list as a starlark.Value
```
timothyg-stripe:
```suggestion
// listValueToStarlark wraps a protobuf list as a starlark.Value.
```

go/protomodule/type_conversions.go
```diff
@@ -194,6 +167,73 @@ func valueToStarlark(val protoreflect.Value, fieldDesc protoreflect.FieldDescrip
...
+		starlarkValue, err := scalarValueToStarlark(listVal.Get(i), fieldDesc)
+		if err != nil {
+			return starlark.None, err
+		}
+		out.Append(starlarkValue)
+	}
+	return out, nil
+}
+
+// Wrap a protobuf map as a starlark.Value
```
timothyg-stripe:
```suggestion
// mapValueToStarlark wraps a protobuf map as a starlark.Value.
```

go/protomodule/type_conversions.go
```diff
@@ -194,6 +167,73 @@ func valueToStarlark(val protoreflect.Value, fieldDesc protoreflect.FieldDescrip
...
+	})
+	if rangeErr != nil {
+		return starlark.None, rangeErr
+	}
+
+	// Attempt to sort the map by key to ensure that the result is "as
+	// deterministic as possible".  Not all starlark values are
+	// comparable. If that is the case, the ordering will be arbitrary.
+	// I believe that *protoMap only supports scalar keys, in which case
+	// the error condition should never be hit.
```
timothyg-stripe, line 223: Not all Starlark values are comparable, but I do believe all (valid) [`protoreflect.MapKey`s](https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect#MapKey) are comparable. Why not sort those instead? I'm pretty sure you can just use `k.Interface()` as the sort key in Go.

timothyg-stripe (commented):
Looks good generally. Could you write some unit tests?

go/protomodule/type_conversions.go
ryanpbrewster: What if I have a map that has some string keys and some float keys? That seems valid, but I don't have an easy way to compare the bare Go values.

timothyg-stripe: Mixed key types are not allowed by protobuf, nor by protoreflect.Map. The godoc for [Map](https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect#Map) says

> The entry [MapKey](https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect#MapKey) type is determined by [FieldDescriptor.MapKey](https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect#FieldDescriptor.MapKey).Kind.

timothyg-stripe: Hmm, it doesn't seem like Go provides an easy way to compare two `any` values through reflection. Never mind – starlark.Compare is probably the best we have. Even so, we should be pretty certain that the error condition won't be a problem. (e.g., protobuf doesn't let map keys be floats)

ryanpbrewster:
The test I added is flaky prior to this change and reliably passes with the change:
```
//go/protomodule:protomodule_test                                        FAILED in 8 out of 10 in 1.0s
```
vs
```
//go/protomodule:protomodule_test                                        PASSED in 1.1s
  Stats over 10 runs: max = 1.1s, min = 0.3s, avg = 0.4s, dev = 0.2s
```

TimothyGu (approved):

ryanpbrewster:
Does your stripe account have write access?

timothyg-stripe (approved):

timothyg-stripe:
@ryanpbrewster thanks for adding the test!

PR #125, opened 2025-04-03
Add `skycfg.WithPositionalArgs`

sushain-stripe:


skycfg.go
```diff
@@ -29,18 +29,17 @@ import (
 	"strings"
 	"time"
 
+	"github.com/stripe/skycfg/go/assertmodule"
```
timothyg-stripe: I think you didn't have the right flags set in `goimports`

timothyg-stripe (dismissed):

skycfg.go
sushain-stripe: Bleh, yeah, VSCode had fun. I'll undo this.

.bazelversion
timothyg-stripe, line 1: Is this intentional?

timothyg-stripe: Honestly I don't understand why this repo uses Bazel at all

sushain-stripe: Yes, it matches what CI uses and makes it so that Bazelisk works out of the box.

> Honestly I don't understand why this repo uses Bazel at all

Heh.

timothyg-stripe (dismissed):

timothyg-stripe (approved):

PR #127, opened 2025-04-04
Upgrade Bazel, rulesets and Go SDK

sushain97:


timothyg-stripe (dismissed):
Do you want to update go.mod too, so that we can actually use newer Go language features?

sushain-stripe:
> Do you want to update go.mod too, so that we can actually use newer Go language features?

Good idea

timothyg-stripe (approved):

PR #121, opened 2024-06-27
Support custom error message in assertion module

LINKIWI:
This PR proposes the addition of an optional secondary argument to `ctx.assert(...)` which enables unit test callsites to specify a custom assertion failure error message.

This looks like, for example:

```starlark
def test_foo(ctx):
    ctx.assert(False)

def test_foo_message(ctx):
    ctx.assert(False, "foo error message")
```

Executing these tests results in:

```
[foo.sky:2:15] assertion failed
...
```

```
[foo.sky:4:15] assertion failed: foo error message
...
```

I added corresponding unit test coverage.

```bash
$ bazelisk test --cache_test_results=no //go/...
INFO: Invocation ID: bc5157d2-13f6-4d45-84b3-42aae65fc926
INFO: Analyzed 10 targets (0 packages loaded, 137 targets configured).
INFO: Found 5 targets and 5 test targets...
INFO: Elapsed time: 2.453s, Critical Path: 0.51s
INFO: 6 processes: 1 internal, 5 processwrapper-sandbox.
INFO: Build completed successfully, 6 total actions
//go/assertmodule:assertmodule_test                                      PASSED in 0.3s
//go/hashmodule:hashmodule_test                                          PASSED in 0.4s
//go/protomodule:protomodule_test                                        PASSED in 0.5s
//go/urlmodule:urlmodule_test                                            PASSED in 0.4s
//go/yamlmodule:yamlmodule_test                                          PASSED in 0.3s
```

PR #130, opened 2025-04-07
Additionally support custom error message in binary-argument assertions

LINKIWI:
This is an extension to #121, which allows specification of a custom assertion error message. #121 supports specifying a message for unary assertions:

```starlark
t.assert(true, "foo")
```

This PR additionally extends the same functionality to support binary assertions:

```starlark
t.assert.equal(1, 1, "foo")
```

Minimal end-to-end example:

```starlark
# foo.sky

def test_foo(ctx):
    ctx.assert.equal(0, 0, "assert1")
    ctx.assert.equal(0, 1, "assert2")
```

```go
// main.go

package main

import (
	"context"
	"fmt"

	"github.com/stripe/skycfg"
)

func main() {
	cfg, err := skycfg.Load(context.Background(), "foo.sky")
	if err != nil {
		panic(err)
	}

	for _, t := range cfg.Tests() {
		result, err := t.Run(context.Background())
		if err != nil {
			panic(err)
		}

		fmt.Println(result.Failure)
	}
}

```

Example output:

```
[foo.sky:3:21] assertion failed: assert2
Traceback (most recent call last):
  foo.sky:3:21: in test_foo
```

### DIFF
--- a/go/assertmodule/assert_test.go
+++ b/go/assertmodule/assert_test.go
@@ -103,7 +103,7 @@ func TestUnaryAsserts(t *testing.T) {
 
 		if testCase.msg != "" {
 			cmd = fmt.Sprintf(
-				`t.assert(%s, "%s")`,
+				`t.assert(%s, %q)`,
 				testCase.val,
 				testCase.msg,
 			)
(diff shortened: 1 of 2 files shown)


go/assertmodule/assert_test.go
```diff
@@ -102,7 +103,7 @@ func TestUnaryAsserts(t *testing.T) {
 
 		if testCase.msg != "" {
 			cmd = fmt.Sprintf(
-				`t.assert(%s, %q)`,
+				`t.assert(%s, "%s")`,
```
sushain-stripe: Mind using `%q` here and below?

sushain-stripe (changes requested):
Looks pretty good. One minor tweak, please

sushain-stripe (approved):

PR #135, opened 2025-04-16
Add support for caching common load()s across different skycfg.Load() calls

timothyg-stripe:


export_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+// This _test.go file exists in package skycfg to expose certain helpers to tests,
+// which live in package skycfg_test.
+// See https://github.com/golang/go/issues/56995 on how this works.
+
+package skycfg
+
+import "sort"
+
+// KeysForTestOnly is a test helper that returns the keys in the [LoadCache].
```
sushain-stripe, line 25: Can we not export this symbol if it's only for tests?

skycfg.go
```diff
@@ -338,33 +384,41 @@ func loadImpl(ctx context.Context, opts *loadOptions, filename string) (starlark
...
 		}
+		loadCalled[modulePath] = struct{}{}
+
 		moduleSource, err := reader.ReadFile(ctx, modulePath)
 		if err != nil {
-			cache[modulePath] = &cacheEntry{nil, err}
-			return nil, err
+			// Make sure to use the existing value in the cache if it already exists.
+			// This can happen if there are two ExecFile calls happening concurrently.
+			// This ensures consistency in case the other load succeeded.
```
sushain-stripe, line 406: Good catch.

sushain-stripe (approved):

export_test.go
timothyg-stripe: It's only exported to tests – notice this file is a `_test.go` file.

sushain-stripe: Ahhh, that's why you did this. OK, I was confused.

PR #136, opened 2025-06-12
Fix nondeterministic oneof initialization

### LINKED ISSUE #103: Non-determinism with oneofs
Here's a reproduction of the issue I'm hitting:

```
package main

import (
	"context"
	"fmt"

	"github.com/stripe/skycfg"
)

func main() {
	ctx := context.Background()
	config, err := skycfg.Load(ctx, "hello.sky")
	if err != nil { panic(err) }
	messages, err := config.Main(ctx)
	if err != nil { panic(err) }
	for _, msg := range messages {
		fmt.Printf("%s\n", msg.String())
	}
}
```
```
# hello.sky
pb = proto.package("google.protobuf")

def main(ctx):
  return [pb.Value(
    bool_value = True,
    string_value = "foo"
  )]
```
Note, `Value` contains a oneof - https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/struct.proto#L62
Repeatedly invocating the code now returns non-deterministic results:
```
$ for x in {1..5}; do ./test-skycfg; done
string_value:"foo"
bool_value:true
bool_value:true
string_value:"foo"
bool_value:true
```
Looks like the non-determinism is in part due to that fact that the fields of the returned proto object are stored internally in a go `map[string]starlark.Value` which has non-deterministic ordering, and the oneof field that is processed last wins - https://github.com/stripe/skycfg/blob/trunk/go/protomodule/protomodule_message.go#L279

Looks like the intent was to mimic the behavior in `dynamicpb` - https://github.com/stripe/skycfg/pull/90/files#r692217496

I decided to compare this behavior with pure python (as it is also dynamically typed and closely resembles starlark), and although you can define something similar:
```
> import google.protobuf.struct_pb2 as pb
> print(pb.Value(string_value="foo", bool_value=True))
bool_value: true  # deterministic
```
The output is deterministic (I suspect because python 3.7 introduced dictionaries that maintained insertion order https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6)

Are the authors open to adding determinism here? This can result in hard to reason about configurations due to inadvertent user error. One naive solution would be to sort fields by name to ensure some type of deterministic ordering.

timothyg-stripe:
With an expression like `proto.package("google.protobuf").Value(string_value="foo", bool_value=True)`, previously it was nondeterministic whether `string_value` or `bool_value` is populated. After this PR, the last kwarg always wins, similar to the protobuf Python.

With the test case added in this PR, `go test -count 50 ./go/protomodule` previously failed regularly, but afterwards succeeds 100% of the time.

Fixes #103.

### DIFF
--- a/go/protomodule/protomodule_message_test.go
+++ b/go/protomodule/protomodule_message_test.go
@@ -110,7 +110,8 @@ func TestMessageV2(t *testing.T) {
 		),
 		f_toplevel_enum = proto.package("skycfg.test_proto").ToplevelEnumV2.TOPLEVEL_ENUM_V2_B,
 		f_nested_enum = proto.package("skycfg.test_proto").MessageV2.NestedEnum.NESTED_ENUM_B,
-		f_oneof_a = "string in oneof",
+		f_oneof_a = "string in oneof but overridden",
+		f_oneof_b = "string in oneof",
 		f_bytes = "also some string",
 
 		# Autoboxed wrappers
@@ -158,7 +159,7 @@ func TestMessageV2(t *testing.T) {
 		},
 		FToplevelEnum: pb.ToplevelEnumV2_TOPLEVEL_ENUM_V2_B.Enum(),
 		FNestedEnum:   pb.MessageV2_NESTED_ENUM_B.Enum(),
-		FOneof:        &pb.MessageV2_FOneofA{"string in oneof"},
+		FOneof:        &pb.MessageV2_FOneofB{FOneofB: "string in oneof"},
 		FBytes:        []byte("also some string"),
 		F_BoolValue:   &wrapperspb.BoolValue{Value: true},
 		F_StringValue: &wrapperspb.StringValue{Value: "something"},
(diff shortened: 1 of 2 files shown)


go/protomodule/protomodule_message_test.go
```diff
@@ -158,7 +159,7 @@ func TestMessageV2(t *testing.T) {
 		},
 		FToplevelEnum: pb.ToplevelEnumV2_TOPLEVEL_ENUM_V2_B.Enum(),
 		FNestedEnum:   pb.MessageV2_NESTED_ENUM_B.Enum(),
-		FOneof:        &pb.MessageV2_FOneofA{"string in oneof"},
+		FOneof:        &pb.MessageV2_FOneofB{FOneofB: "string in oneof"},
```
sushain-stripe, line 162: I assume these changes were all to silence IDE warnings?

go/protomodule/protomodule_message_type.go
```diff
@@ -145,7 +145,9 @@ func (t *protoMessageType) CallInternal(
 	if err != nil {
 		return nil, err
 	}
-	for attrName, starlarkValue := range parsedKwargs {
+	for _, kwarg := range kwargs {
```
sushain-stripe, line 153: This is worth a comment, it doesn't immediately jump out as wrong to refactor.

sushain-stripe (dismissed):

go/protomodule/protomodule_message_test.go
timothyg-stripe: yup

go/protomodule/protomodule_message_type.go
timothyg-stripe: fixed

sushain-stripe (approved):

PR #138, opened 2025-06-12
Upgrade starlark and allow customizing syntax.FileOptions

### LINKED ISSUE #137: skycfg should allow changing starlark/syntax.FileOptions
[syntax.FileOptions](https://pkg.go.dev/go.starlark.net@v0.0.0-20250603171236-27fdb1d4744d/syntax#FileOptions) is the modern way to change Starlark language options, like whether `set` is allowed. We should allow users to customize this value.

timothyg-stripe:
Fixes https://github.com/stripe/skycfg/issues/137.


go/yamlmodule/yamlmodule.go
```diff
@@ -22,26 +22,26 @@ import (
 	"fmt"
 	"reflect"
 
+	starlarkjson "go.starlark.net/lib/json"
 	"go.starlark.net/starlark"
-	"go.starlark.net/starlarkjson"
```
timothyg-stripe, line 26: Drive-by: starlarkjson is deprecated in favor of lib/json. They do the same thing.

sushain-stripe (approved):

PR #139, opened 2025-06-13
Add FSFileReader

timothyg-stripe:
Add an implementation of [FileReader](https://pkg.go.dev/github.com/stripe/skycfg#FileReader) that reads from a [io/fs.FS](https://pkg.go.dev/io/fs#FS). This is a version of what we've been using internally.

### DIFF
--- a/skycfg.go
+++ b/skycfg.go
@@ -22,7 +22,7 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"io/ioutil"
+	"io/fs"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


skycfg.go
```diff
@@ -90,7 +92,28 @@ func (r *localFileReader) Resolve(ctx context.Context, name, fromPath string) (s
...
+	return &fsFileReader{fsys: fsys}
+}
+
+func (r *fsFileReader) Resolve(ctx context.Context, name, fromPath string) (string, error) {
+	return path.Clean(name), nil
+}
+
+func (r *fsFileReader) ReadFile(ctx context.Context, path string) ([]byte, error) {
+	return fs.ReadFile(r.fsys, path)
 }
```
sushain-stripe: Thoughts on sticking the readers in a separate file?

sushain-stripe (dismissed):

skycfg.go
timothyg-stripe: sure

sushain-stripe (approved):

PR #140, opened 2025-06-18
go/protomodule: add go1.23 iter support

timothyg-stripe:
Add `(*protoMap).Entries` and `(*protoList).Elements`, which return `iter.Seq2` and `iter.Seq` respectively, to mirror https://github.com/google/starlark-go/pull/551.


go/protomodule/protomodule_test.go
```diff
@@ -34,10 +33,6 @@ import (
 	pb "github.com/stripe/skycfg/internal/test_proto"
 )
 
-func init() {
-	resolve.AllowFloat = true
```
timothyg-stripe, line 38: Drive-by: this is true by default now.

sushain-stripe (approved):

PR #143, opened 2025-06-23
debugger: Upgrade Starlark to remove getPredeclared hack

timothyg-stripe:
https://github.com/google/starlark-go/issues/594 was recently fixed. Instead of rolling our own hacky `getPredeclared(starFunc)`, we can now say `starFunc.Module().Predeclared()`.


debugger.test
timothyg-stripe, line 1: oops, sorry for adding this

sushain-stripe: rip, my bad for not noticing.

sushain-stripe (approved):
🎉

PR #145, opened 2025-11-09
protomodule: fix freeze

timothyg-stripe:
Skycfg previously allowed something like
```starlark
frozen_msg.r_string.append("foo")
```
to run, where `frozen_msg` is a frozen protobuf message object and `r_string` is a `repeated string` field.

In practice, this is benign because a new list object is created every time `frozen_msg.r_string` is accessed, and the message itself is unchanged. But it's certainly unexpected behavior.

### DIFF
--- a/go/protomodule/protomodule_message_test.go
+++ b/go/protomodule/protomodule_message_test.go
@@ -32,6 +32,8 @@ import (
 )
 
 func TestMessageAttrNames(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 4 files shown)


go/protomodule/protomodule_message_test.go
```diff
@@ -32,6 +32,8 @@ import (
 )
 
 func TestMessageAttrNames(t *testing.T) {
+	t.Parallel()
```
sushain-stripe, line 35: Thanks

sushain-stripe (approved):

PR #146, opened 2025-11-10
protomodule: performance optimizations

timothyg-stripe:
```
❯ go install golang.org/x/perf/cmd/benchstat@latest
❯ go test -bench=. -count=10 ./go/protomodule >before|after.txt
❯ ~/go/bin/benchstat before.txt after.txt 
goos: darwin
goarch: arm64
pkg: github.com/stripe/skycfg/go/protomodule
cpu: Apple M1 Max
                                        │  before.txt  │              after.txt               │
                                        │    sec/op    │    sec/op     vs base                │
NewMessage_empty-10                        507.2n ± 6%   457.6n ± 10%   -9.77% (p=0.002 n=10)
NewMessage_full-10                         14.99µ ± 5%   10.21µ ±  9%  -31.94% (p=0.000 n=10)
Message_Attr_unpopulatedScalar-10          27.00n ± 3%   31.99n ±  2%  +18.46% (p=0.000 n=10)
Message_Attr_populatedScalar-10            9.104n ± 5%   8.797n ±  2%        ~ (p=0.052 n=10)
Message_Attr_unpopulatedList-10            294.6n ± 3%   155.5n ± 10%  -47.23% (p=0.000 n=10)
Message_Attr_populatedList-10              8.719n ± 4%   8.457n ±  3%        ~ (p=0.165 n=10)
Message_Attr_unpopulatedLastItem-10       259.25n ± 4%   36.44n ±  1%  -85.94% (p=0.000 n=10)
Message_Attr_populatedLastItem-10          15.24n ± 3%   15.05n ±  2%        ~ (p=0.382 n=10)
Message_SetField_unpopulatedInt32-10       41.20n ± 4%   41.73n ±  2%        ~ (p=0.247 n=10)
Message_SetField_populatedInt32-10         40.85n ± 4%   41.79n ±  3%   +2.30% (p=0.023 n=10)
Message_SetField_unpopulatedLastItem-10   264.70n ± 4%   48.40n ±  4%  -81.72% (p=0.000 n=10)
Message_SetField_populatedLastItem-10     266.85n ± 4%   48.77n ±  3%  -81.72% (p=0.000 n=10)
geomean                                    107.0n        62.99n        -41.14%
```

* `NewMessage_empty-10` got faster mostly because of the replacement of `proto.Clone` + `proto.Reset` with `msg.ProtoReflect().New()`.
* `NewMessage_full-10` got faster due to the removal of `proto.Clone` and the use of `appendUnchecked` + `setKeyUnchecked`.
* `Message_Attr_unpopulatedScalar-10` and `Message_SetField_*Int32-10` got a little slower, primarily because those fields are near the beginning of the message descriptor, so the current linear scan in `getFieldDescriptor` ends up being faster than a `map` lookup. However, the same change made `BenchmarkMessage_Attr_unpopulatedList-10`, `Message_Attr_unpopulatedLastItem-10`, and `Message_SetField_*LastItem-10` **much** faster, so it's worth it overall

### DIFF
--- a/go/protomodule/bench_test.go
+++ b/go/protomodule/bench_test.go
@@ -1,0 +1,255 @@
+package protomodule
+
+import (
+	"testing"
+
+	pb "github.com/stripe/skycfg/internal/test_proto"
+	"go.starlark.net/starlark"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+)
+
+var (
+	benchmarkNewMessage_empty_sink *protoMessage
+	benchmarkNewMessage_empty_err  error
+)
+
+func BenchmarkNewMessage_empty(b *testing.B) {
+	wantMsg := &pb.MessageV2{}
+	benchmarkNewMessage_empty_sink, benchmarkNewMessage_empty_err = NewMessage(wantMsg)
+	b.ResetTimer()
+	for range b.N {
+		benchmarkNewMessage_empty_sink, benchmarkNewMessage_empty_err = NewMessage(wantMsg)
+	}
+}
+
+var (
+	benchmarkNewMessage_full_sink *protoMessage
+	benchmarkNewMessage_full_err  error
+)
+
+func BenchmarkNewMessage_full(b *testing.B) {
+	wantMsg := &pb.MessageV2{
+		FInt32:   proto.Int32(1010),
+		FInt64:   proto.Int64(1020),
+		FUint32:  proto.Uint32(1030),
+		FUint64:  proto.Uint64(1040),
+		FFloat32: proto.Float32(10.50),
+		FFloat64: proto.Float64(10.60),
+		FString:  proto.String("some string"),
+		FBool:    proto.Bool(true),
+		FSubmsg: &pb.MessageV2{
+			FString: proto.String("string in submsg"),
+		},
+		RString: []string{"r_string1", "r_string2"},
+		RSubmsg: []*pb.MessageV2{{
+			FString: proto.String("string in r_submsg"),
+		}},
+		MapString: map[string]string{
+			"map_string key": "map_string val",
+		},
+		MapSubmsg: map[string]*pb.MessageV2{
+			"map_submsg key": {
+				FString: proto.String("map_submsg val"),
+			},
+		},
+		FNestedSubmsg: &pb.MessageV2_NestedMessage{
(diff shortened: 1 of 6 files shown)


go/protomodule/protomodule_message.go
```diff
@@ -90,11 +87,10 @@ func AsProtoMessage(v starlark.Value) (proto.Message, bool) {
 // proto.Message through AsProtoMessage. Any fields set to starlark.None or the
 // default field value will be ignored when returning to a protobuf.Message
 type protoMessage struct {
-	// A copy of the underlying is stored so AsProtoMessage can construct a new object
```
timothyg-stripe, line 93: A previous implementation may have reflected every mutation back to `msg`, but that's no longer the case: `msg` is always empty. Make that explicit by renaming the field, and also store a protoreflect.Message since that ends up being used more often.

go/protomodule/bench_test.go
```diff
@@ -0,0 +1,255 @@
...
+
+	for range b.N {
+		benchmarkMessage_Attr_unpopulatedLastItem_sink, benchmarkMessage_Attr_unpopulatedLastItem_err = msg.Attr("safe_")
+	}
+}
+
+var (
+	benchmarkMessage_Attr_populatedLastItem_sink starlark.Value
+	benchmarkMessage_Attr_populatedLastItem_err  error
+)
```
sushain-stripe, line 175: Hm, why are these consistently declared in the module scope rather than function scope? I'm not seeing them accessed outside of the function.

sushain-stripe (approved):
🎉

go/protomodule/bench_test.go
timothyg-stripe: They are here as sinks to prevent the compiler from optimizing out anything in the loop. We are not yet on a new enough version of Go that supports [`for b.Loop()`](https://pkg.go.dev/testing#B.Loop), which has special compiler support to not optimize out seemingly dead code.

sushain-stripe: Ah, that seems like it definitely needs a comment :) Unless it's a common idiom?

timothyg-stripe: TBH I don't know if it's even necessary here because I doubt the Go compiler would compile out any one of these functions. I'm hoping to upgrade to b.Loop soon and not worry about this.

PR #147, opened 2025-11-13
parallelmodule: Create

timothyg-stripe:
Create a module for running Starlark functions in parallel. Starlark is designed to support safe parallel evaluation, so why not make some functions for doing that.

It's not loaded by skycfg by default, because many configuration-based use cases may not want to allow such a powerful feature.

Also fix macOS 26 compatibility by setting the `-B gobuildid` linker flag (https://go.dev/issues/68678); Go 1.24+ set this by default. Without the flag, I get
```
dyld[32468]: missing LC_UUID load command in /private/var/tmp/_bazel/da12fa8050f39c66683875ec236a6a61/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/external/gazelle+/cmd/gazelle/gazelle_/gazelle
dyld[32468]: missing LC_UUID load command
/private/var/tmp/_bazel/da12fa8050f39c66683875ec236a6a61/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/gazelle: line 112: 32468 Abort trap: 6           "$gazelle_path" "${ARGS[@]}"
```

### DIFF
--- a/go/parallelmodule/parallelmodule.go
+++ b/go/parallelmodule/parallelmodule.go
@@ -1,0 +1,194 @@
+// Package parallelmodule contains Starlark functions for running code in parallel.
+// Unlike other Starlark modules in skycfg, this module is not loaded by skycfg by default.
+package parallelmodule
+
+import (
+	"fmt"
+
+	"go.starlark.net/starlark"
+	"go.starlark.net/starlarkstruct"
+	"go.starlark.net/syntax"
+	"golang.org/x/sync/errgroup"
+)
+
+// NewModule returns a Starlark module of parallel functions.
+//
+//	parallel = module(
+//	    map,
+//	)
+//
+// def map(function, iterable):
+//
+// Runs function on each element of iterable in parallel.
+// Returns a list containing the result of function, in the same order as the
+// original iterable.
+// If any call to function fails, parallel.map fails as well.
+// If multiple invocations fail, the error is chosen arbitrarily.
+//
+// To guarantee safety, the following measures are taken:
+//   - The function and each element of iterable are frozen.
+//   - map cannot be called during module initialization.
+//     See https://github.com/google/starlark-go/issues/623.
+//   - map detects and forbids recursion, even if Starlark is configured to
+//     allow recursion.
+var Module = &starlarkstruct.Module{
+	Name: "parallel",
+	Members: starlark.StringDict{
+		"map": starlark.NewBuiltin("parallel.map", parMap),
+	},
+}
+
+// makeTrampoline creates a function appropriate for [starlark.NewBuiltin] that
+// just calls the provided callable with args and wargs.
+func makeTrampoline(callable starlark.Callable) func(_ *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	return func(t *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+		return starlark.Call(t, callable, args, kwargs)
+	}
+}
+
+// callableWithPos is a callable that always has the given position in a stack trace.
+type callableWithPos struct {
+	starlark.Callable
+	pos syntax.Position
+}
+
+func (f callableWithPos) Position() syntax.Position {
+	return f.pos
+}
+
+// replicateStack creates a Starlark callable that, when called, adds the call
+// frames of t to the call stack, and then calls fn.
+func replicateStack(t *starlark.Thread, fn starlark.Callable) starlark.Callable {
+	for depth := range t.CallStackDepth() {
+		fr := t.CallFrame(depth)
+		builtin := starlark.NewBuiltin(fr.Name, makeTrampoline(fn))
+		fn = callableWithPos{builtin, fr.Pos}
+	}
+	return fn
+}
+
+func checkRecursionOrInitialization(t *starlark.Thread, fn *starlark.Builtin) error {
+	seenPos := map[syntax.Position]struct{}{}
+	for depth := range t.CallStackDepth() {
+		fr := t.CallFrame(depth)
+
+		// We cannot guarantee data race freedom during module initialization,
+		// because freezing a function does not prevent a global from being
+		// modified or reassigned (if -globalreassign).
+		// See https://github.com/google/starlark-go/issues/623.
+		//
+		// Merely freezing the mapper's globals is **not** sufficient:
+		// the iterable can also contain functions with access to globals.
+		// See TestMap_mutate_global_during_toplevel2.
+		//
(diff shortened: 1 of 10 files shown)


go/protomodule/BUILD
```diff
@@ -34,6 +34,7 @@ go_library(
 go_test(
     name = "protomodule_test",
     srcs = [
+        "bench_test.go",
```
timothyg-stripe, line 37: I forgot to run Gazelle when I added this file…

tools/rules_go_macos.patch
```diff
@@ -0,0 +1,14 @@
...
++++ b/go/private/go_toolchain.bzl	2025-11-12 21:48:21.186200804 -0800
+@@ -111,6 +111,11 @@
+             cgo_link_flags.extend(["-shared", "-Wl,-all_load"])
+         if host_goos == "linux":
+             cgo_link_flags.append("-Wl,-whole-archive")
++        if p.goos == "darwin":
++            # Add -B gobuildid for compatibility with macOS 26.
++            # Only necessary before Go 1.24.
++            # See https://go.dev/issues/68678
++            link_flags.extend(["-B", "gobuildid"])
```
timothyg-stripe, line 11: rules_go doesn't have an easier way to modify this other than creating your own `go_toolchain`s… tsk tsk. rules_rust is a lot more configurable in this regard.

go/parallelmodule/parallelmodule.go
```diff
@@ -0,0 +1,194 @@
...
+		// Merely freezing the mapper's globals is **not** sufficient:
+		// the iterable can also contain functions with access to globals.
+		// See TestMap_mutate_global_during_toplevel2.
+		//
+		// On the other hand, if a module is fully initialized, race freedom is
+		// guaranteed:
+		//  1. Globals can never be reassigned by non-top-level Starlark functions.
+		//  2. Global values must have been frozen already at the end of module
+		//     initialization.
+		if fr.Name == "<toplevel>" {
```
sushain-stripe, line 89: A bit unfortunate that there's no constant for this but I assume you checked (similar comments below)

go/parallelmodule/parallelmodule.go
```diff
@@ -0,0 +1,194 @@
...
+// To guarantee safety, the following measures are taken:
+//   - The function and each element of iterable are frozen.
+//   - map cannot be called during module initialization.
+//     See https://github.com/google/starlark-go/issues/623.
+//   - map detects and forbids recursion, even if Starlark is configured to
+//     allow recursion.
+var Module = &starlarkstruct.Module{
+	Name: "parallel",
+	Members: starlark.StringDict{
+		"map": starlark.NewBuiltin("parallel.map", parMap),
```
sushain-stripe, line 37: I could imagine wanting bounded concurrency here but that seems fine to add whenever we find a need for it (we could use a kwarg).

go/protomodule/BUILD
sushain-stripe: we should _probably_ run it in GH actions?

go/parallelmodule/parallelmodule.go
```diff
@@ -0,0 +1,194 @@
...
+	return starlark.NewList(vals), nil
+}
+
+// toFrozenSlice converts an iterable to a slice of frozen values.
+func toFrozenSlice(seq starlark.Iterable) (out []starlark.Value) {
+	iter := seq.Iterate()
+	defer iter.Done()
+	var elem starlark.Value
+	for iter.Next(&elem) {
+		elem.Freeze()
```
sushain-stripe, line 190: Hm, does this break later mutation of the inputs? I think that's fair given your function description but perhaps a bit unpleasant…

go/parallelmodule/parallelmodule_test.go
```diff
@@ -0,0 +1,504 @@
...
+		assert.Equal(t, expectedKwargs, kwargs)
+		callStack = thread.CallStack()
+		return starlark.None, nil
+	})
+
+	filename := "foobar"
+	pos := syntax.MakePosition(&filename, 42, 1)
+	callable := callableWithPos{builtin, pos}
+	_, err := starlark.Call(new(starlark.Thread), callable, expectedArgs, expectedKwargs)
+	if assert.NoError(t, err) {
```
sushain-stripe, line 34: Hm, would it not reduce indentation to use `require` here and below?

go/parallelmodule/parallelmodule_test.go
```diff
@@ -0,0 +1,504 @@
...
+	elements := []starlark.Value{
+		starlark.String("foo"),
+		starlark.String("bar"),
+		starlark.String("baz"),
+	}
+
+	s := starlark.NewSet(len(elements))
+	for _, val := range elements {
+		err := s.Insert(val)
+		assert.NoError(t, err)
```
sushain-stripe:
```suggestion
		require.NoError(t, err)
```

sushain-stripe (dismissed):
Great work with all those tests! Sorry for the delayed review

go/parallelmodule/parallelmodule.go
timothyg-stripe: Yeah, I think there's a lot more we could do here but didn't want to spend too long designing another threading/threadpool/etc. API lol

timothyg-stripe: I was thinking the skycfg host should have a way to control how much parallelism an skycfg program can use.

go/parallelmodule/parallelmodule.go
timothyg-stripe: This is important to guarantee safety. Imagine what would happen without this freeze for a program like:
```starlark
the_same_list_but_six_times = [[]] * 6
# The above is equivalent to
# l = []
# the_same_list_but_six_times = [l, l, l, l, l, l]

def append_fn(l):
    l.append(1)
parallel.map(append_fn, the_same_list_but_six_times)
```

Chaos.

go/parallelmodule/parallelmodule_test.go
timothyg-stripe: It would, but I've taken a liking to this style.

go/parallelmodule/parallelmodule.go
sushain-stripe: Oh certainly, I agree it's required! I think the alternative would be copying the inputs but that is perhaps worse?

timothyg-stripe: Yeah, I think that's worse.

sushain-stripe (approved):

go/parallelmodule/parallelmodule.go
timothyg-stripe: Starlark doesn't really have a well-defined copying operation, unlike, say, [JavaScript](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm). So it's best to leave that to the caller.

PR #150, opened 2026-06-10
allow parallelmodule.map to take a new `limit` arg

sarahan-stripe:
Adds a new optional arg `limit` to parallelmodule.map.
This allows callers of parallelmodule.map to manage the number of active goroutines in the errgroup. It defaults to `-1`, or "no limit".

### DIFF
--- a/go/parallelmodule/parallelmodule.go
+++ b/go/parallelmodule/parallelmodule.go
@@ -24,7 +24,7 @@ import (
 // original iterable.
 // If any call to function fails, parallel.map fails as well.
 // If multiple invocations fail, the error is chosen arbitrarily.
-// If limit is provided and positive, at most limit goroutines run concurrently.
+// If limit is provided and nonnegative, at most limit goroutines run concurrently.
 //
 // To guarantee safety, the following measures are taken:
 //   - The function and each element of iterable are frozen.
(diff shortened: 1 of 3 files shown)


go/parallelmodule/parallelmodule.go
```diff
@@ -17,13 +17,14 @@ import (
...
 //
-// def map(function, iterable):
+// def map(function, iterable, limit=-1):
 //
 // Runs function on each element of iterable in parallel.
 // Returns a list containing the result of function, in the same order as the
 // original iterable.
 // If any call to function fails, parallel.map fails as well.
 // If multiple invocations fail, the error is chosen arbitrarily.
+// If limit is provided and non-negative, at most `limit` goroutines run concurrently.
```
timothyg-stripe: If `limit = 0`, this function will infinite loop, right? That's pretty error-prone. We might as well make `0` the same as `-1`.

Also for documentation's sake, let's not expose words like "goroutines". The docs are supposed to be read for people who write skycfg, not Go.

Lastly, please edit https://github.com/sarahan-stripe/skycfg/blob/c7cd15d810ec87d762a28f342efb23fa97d714da/docs/modules.asciidoc#parallel too.

timothyg-stripe (changes requested):

timothyg-stripe (approved):

